PR #139, opened 2020-05-15
One NodeShared per process

joxoby:
https://github.com/orgs/ignitionrobotics/projects/3#card-38255453

### DIFF
--- a/src/NodeShared.cc
+++ b/src/NodeShared.cc
@@ -32,6 +32,8 @@
 #include <thread>
 #include <vector>
 
+#include <unistd.h>
+
 // TODO(anyone): Remove after fixing the warnings.
 #ifdef _MSC_VER
 #pragma warning(push, 0)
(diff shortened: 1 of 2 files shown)


src/NodeShared.cc
```diff
@@ -139,12 +138,17 @@ NodeShared *NodeShared::Instance()
   // application exits. We may want to consider a more elegant solution in
   // the future. The zsys_shutdown() function in the czmq library may be able
   // to provide some inspiration for solving this more cleanly.
-  static NodeShared *instance = new NodeShared();
+  static NodeShared *instance = nullptr;
+
+  // Create a new instance of NodeShared if the process has changed
+  // (maybe after fork?) so the ZMQ context is not shared between different
+  // processes.
+  if (instance == nullptr || instance->pUuid != Uuid().ToString())
```
joxoby, line 146: We may want to make this `Uuid` comparison faster

caguero:
Thanks for the contribution. I've been doing some testing and I have two major concerns:

joxoby:
Hi Carlos,
1. You're right, it was a confusion of mine regarding `Uuid`. What I meant to codify was something like
```
if (instance == nullptr || instance->pid != getpid())
```
2. We can easily add some `lock_guards`. WDYT?

caguero:
> instance == nullptr || instance->pid != getpid()

Do you mind to update this pull request with a working prototype? It'll help to wrap my head around it. Thanks!

joxoby:
With the last change, the `ignition-transport` test are passing but it does cause problems with `ign-gazebo`. Will update.

caguero:
I'll decline this pull request for now until there's a better patch that keeps all tests and examples working. Feel free to reopen if that's the case.

PR #149, opened 2020-07-10
Update deps for Dome: use msgs6

j-rivero:
As part of the works of https://github.com/ignition-tooling/gazebodistro/pull/5 bump the use of msgs to major version 6

### DIFF
--- a/.github/workflows/ci-bionic.yml
+++ b/.github/workflows/ci-bionic.yml
@@ -13,5 +13,5 @@ jobs:
         id: ci
         uses: ignition-tooling/ubuntu-bionic-ci-action@master
         with:
-          apt-dependencies: 'pkg-config libprotobuf-dev protobuf-compiler libprotoc-dev libzmq3-dev uuid-dev libsqlite3-dev libignition-cmake2-dev libignition-math6-dev libignition-msgs5-dev libignition-tools-dev'
+          apt-dependencies: 'pkg-config libprotobuf-dev protobuf-compiler libprotoc-dev libzmq3-dev uuid-dev libsqlite3-dev libignition-cmake2-dev libignition-math6-dev libignition-msgs6-dev libignition-tools-dev'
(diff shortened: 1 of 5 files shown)


caguero:
Should we also remove `bitbucket-pipelines`?

j-rivero:
> How about updating `.gitbug/workflows/ci-bionic.yaml`? Otherwise I think CI will break.

I would like to avoid the duplication of build-from-source dependencies (we already do that in release-tools) since it adds a good amount of overhead to manage changes in all the repositories, at least until we have a common solution in place. We can discuss this within the project, it is a structural problem of having CI code in many different places.

> Should we also remove `bitbucket-pipelines`?

We are still using it to determine the platform to build branches. That functionality needs to be migrated first.

chapulina:
Pending CI fixes:

* Windows is looking for branch `ign-mshs6` when it should be `master`.
* Ubuntu on Jenkins: I just merged https://github.com/ignition-tooling/release-tools/pull/239, which should fix it.
* OSX: I broke off only the `ign-transport` changes into https://github.com/osrf/homebrew-simulation/pull/1065
* Ubuntu on Actions: @j-rivero said he was going to open a PR adding nightlies to the action

j-rivero:
Github actions are reporting green although [it is a false positive](
https://github.com/ignitionrobotics/ign-transport/pull/149/checks?check_run_id=881935176), related to https://github.com/ignition-tooling/ubuntu-bionic-ci-action/pull/11.

In order to make it work properly with nightlies we need https://github.com/ignition-tooling/ubuntu-bionic-ci-action/pull/12 and change this repo to use `gzdev_project_name`.

chapulina:

caguero:
> All Jenkins CI is passing. OSX has 1 test failure and OSX / Windows have warnings. I think we can fix actions later, because all branches will need fixes.
> 
> What do you think, @caguero , can we merge so we can move on with the downstream libs?

Looks good!

PR #153, opened 2020-07-28
Helper function to get a valid topic name

chapulina:
Addresses https://github.com/ignitionrobotics/ign-gazebo/issues/239
Used by https://github.com/ignitionrobotics/ign-sensors/pull/33

Add a helper function to generate a valid topic name by converting spaces to `_` and stripping invalid character combinations.

### DIFF
--- a/src/TopicUtils.cc
+++ b/src/TopicUtils.cc
@@ -15,6 +15,7 @@
  *
 */
 
+#include <regex>
 #include <string>
 
 #include "ignition/transport/TopicUtils.hh"
@@ -153,4 +154,26 @@ bool TopicUtils::DecomposeFullyQualifiedTopic(
   _partition = possiblePartition;
   _namespaceAndTopic = possibleTopic;
   return true;
+}
+
+//////////////////////////////////////////////////
+std::string TopicUtils::AsValidTopic(const std::string &_topic)
+{
+  std::string validTopic{_topic};
+
+  // Substitute spaces with _
+  validTopic = std::regex_replace(validTopic, std::regex(" "), "_");
+
+  // Remove special characters and combinations
+  validTopic = std::regex_replace(validTopic, std::regex("@"), "");
+  validTopic = std::regex_replace(validTopic, std::regex("~"), "");
+  validTopic = std::regex_replace(validTopic, std::regex("//"), "");
+  validTopic = std::regex_replace(validTopic, std::regex(":="), "");
+
+  if (!IsValidTopic(validTopic))
+  {
+    return std::string();
+  }
+
+  return validTopic;
 }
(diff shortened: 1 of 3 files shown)


src/TopicUtils.cc
```diff
@@ -154,3 +155,25 @@ bool TopicUtils::DecomposeFullyQualifiedTopic(
...
+  std::string validTopic{_topic};
+
+  // Substitute spaces with _
+  validTopic = std::regex_replace(validTopic, std::regex(" "), "_");
+
+  // Remove special characters and combinations
+  validTopic = std::regex_replace(validTopic, std::regex("@"), "");
+  validTopic = std::regex_replace(validTopic, std::regex("~"), "");
+  validTopic = std::regex_replace(validTopic, std::regex("//"), "");
+  validTopic = std::regex_replace(validTopic, std::regex(":="), "");
```
iche033, line 171: I think we can merge these regex expressions:

```
validTopic = std::regex_replace(validTopic, std::regex("@|~|//|:="), "");
```

caguero: As a sanity check, can we measure how much time does it take to run this function? I remember seeing awful performance using `std::regex`.

chapulina: I combined all the regex calls in 43678c1.

---

_tl;dr regex performance does look bad, but I think it's acceptable for the current use case_

I did a little performance test comparing the time it takes to run the entire `asValidTopic` test 1000 times, without the `EXPECT` calls, in 2 different ways:

* Without changes to `AsValidTopic`: takes around 0.050 s
* With the regex calls commented out from `AsValidTopic`: takes around 0.0014 s

That does look like a lot. I'm not sure if there's a better alternative though, looping through the string feels prone to error. If all patterns were single characters I think we'd have more options.

For the usage we currently have on `ign-sensors`, I think this is acceptable though. This function is being called only once per sensor when their topic is set. Currently, this is not something that will be called repeatedly.

But let me know if you can think of a better alternative.

src/TopicUtils_TEST.cc
```diff
@@ -242,6 +242,55 @@ TEST(TopicUtilsTest, testFullyQualifiedName)
...
+      "ABC/",
+      "/abc",
+      "/abc/d",
+      "/abc/d/e",
+      "a(bc)d-e_f=h+i.j"
+    })
+  {
+    auto valid = transport::TopicUtils::AsValidTopic(unmodified);
+    EXPECT_EQ(unmodified, valid);
+    EXPECT_TRUE(transport::TopicUtils::IsValidTopic(valid)) << valid;
```
caguero, line 262: I confess that I didn't know that the GTest macros were returning a stream. A good way to debug what failed I guess. :+1:

src/TopicUtils.cc
caguero: It's totally acceptable to me as long as this is not called repeatedly. :+1:

PR #150, opened 2020-07-15
Remove warnings on Homebrew

caguero:
No warnings anymore! 

https://build.osrfoundation.org/job/ignition_transport-ci-pr_any-homebrew-amd64/827/

:green_salad:

Issue #157, opened 2020-07-30
Unable to build on Ubuntu 20.04


pohzhiee:
**Note that this is a bug with ZMQ's ubuntu package, and not ign-transport per se, but it still affects ign-transport nonetheless.**
# Description
The following error messages will be shown during build when using the ign-transport8 branch.
```
/home/pohzhiee/ign-workspace2/src/ign-transport/src/NodeShared.cc:1072:32: error: ‘class zmq::socket_t’ has no member named ‘set’
 1072 |     this->dataPtr->subscriber->set(zmq::sockopt::subscribe, topic);
      |                                ^~~
/home/pohzhiee/ign-workspace2/src/ign-transport/src/NodeShared.cc:1072:41: error: ‘zmq::sockopt’ has not been declared
 1072 |     this->dataPtr->subscriber->set(zmq::sockopt::subscribe, topic);
      
```

# Cause
This is due to the change in this [commit](https://github.com/ignitionrobotics/ign-transport/commit/d4420b83a01d97e5ef1d5326712eb1c1b44de289). Specifically, any addition of the following macro in NodeShared.cc:
```c
 #if (CPPZMQ_VERSION >= 40700)
```

This is due to a bug with the ZMQ versioning in the libzmq3-dev package in ubuntu. Upon installation, the  /usr/include/zmq.hpp will have the following lines:
```c
#define CPPZMQ_VERSION_MAJOR 4
#define CPPZMQ_VERSION_MINOR 7
#define CPPZMQ_VERSION_PATCH 0
```
However, the libzmq shipped with ubuntu 20.04 is 4.3.2 which is reflected in the official ubuntu packages [here](https://packages.ubuntu.com/focal/libzmq3-dev).

# Verification
Run a ubuntu:focal docker container and check that CPPZMQ_VERSION is indeed set to 4.7.0 after installing libzmq3-dev
```sh
docker run -it ubuntu:focal
```
While inside the container
```
apt update && apt install -y libzmq3-dev
grep -i version /usr/include/zmq.hpp
```

# Other notes
The ubuntu bionic libzmq3-dev zmq.hpp does not set version to 4.7.0 so no issues when building with 18.04. This issue only affects 20.04.

# Temporary fixes
Any of the following 2 options will resolve the problem temporarily:
1. Modify /usr/include/zmq.hpp to reflect the correct version
2. Rollback the mentioned commit

mjcarroll:
[...]

PR #159, opened 2020-08-06
Suppress focal-specific warnings

mjcarroll:
Addresses #157 and fixes #150.

Both homebrew and Ubuntu Focal package a `cppzmq` version that is reported as 4.7.0, but the deprecation warnings and available functions are different.  I believe this is due in part to there not being an official 4.7.0 release.

This simply detects if we are building on Ubuntu Focal and adjusts accordingly.


src/Node.cc
```diff
@@ -619,7 +619,7 @@ bool Node::Unsubscribe(const std::string &_topic)
   if (!this->dataPtr->shared->localSubscribers
       .HasSubscriber(fullyQualifiedTopic))
   {
-#if (CPPZMQ_VERSION >= 40700)
+#ifdef IGN_CPPZMQ_POST_4_7_0
```
chapulina, line 622: Homebrew is giving this warning:

`'setsockopt' is deprecated: from 4.7.0, use `set` taking option from zmq::sockopt`

mjcarroll: I'm a bit stumped why this is triggering a warning in `Node.cc`, but not in `NodeShared.cc`, where the same pattern is used.

CMakeLists.txt
```diff
@@ -46,6 +46,19 @@ ign_find_package(IgnProtobuf
 # Find ZeroMQ
 ign_find_package(ZeroMQ VERSION 4 REQUIRED PRIVATE)
 
+if (CMAKE_SYSTEM_NAME MATCHES "Linux")
+  execute_process(COMMAND lsb_release -cs
+    OUTPUT_VARIABLE RELEASE_CODENAME
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_QUIET)
```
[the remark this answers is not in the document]
mjcarroll: Added a warning in 0fbc40f

src/Node.cc
mjcarroll: I was including the incorrect header for the version check, this should be addressed now.

chapulina:
The `Log_TEST` failed on Focal, not sure if it could be caused by this PR.

mjcarroll:
> The Log_TEST failed on Focal, not sure if it could be caused by this PR.

It happens locally for me, as well.  I think because this test doesn't involve zmq, it's more likely a difference between bionic and focal library versions (possibly `sqlite`?)

https://github.com/ignitionrobotics/ign-transport/blob/ign-transport8/log/src/Log_TEST.cc#L250-L259

nkoenig:
> > The Log_TEST failed on Focal, not sure if it could be caused by this PR.
> 
> It happens locally for me, as well. I think because this test doesn't involve zmq, it's more likely a difference between bionic and focal library versions (possibly `sqlite`?)
> 
> https://github.com/ignitionrobotics/ign-transport/blob/ign-transport8/log/src/Log_TEST.cc#L250-L259

Focal outputs a value of 7254000000ns. I'll try adding an #ifdef.

caguero:
> > > The Log_TEST failed on Focal, not sure if it could be caused by this PR.
> > 
> > 
> > It happens locally for me, as well. I think because this test doesn't involve zmq, it's more likely a difference between bionic and focal library versions (possibly `sqlite`?)
> > https://github.com/ignitionrobotics/ign-transport/blob/ign-transport8/log/src/Log_TEST.cc#L250-L259
> 
> Focal outputs a value of 7254000000ns. I'll try adding an #ifdef.

I relaxed the expectation for not having to go back to this test (hopefully) after another Ubuntu release.

caguero (approved):

PR #164, opened 2020-08-26
Prevent empty messages from spamming the console.

nkoenig:
A topic could publish an Int32 or Int64 message with data=0. In this situation the protobuf message has size zero. While this is not a problem for protobuf messages, it is considered an error for SQLite3. 

Added logic to prevent console spamming.


caguero (approved):
That protobuf behavior for default values might be more efficient but it's super annoying. I remember hitting us multiple times.

PR #155, opened 2020-07-30
 7 ➡️ 8 

chapulina:
I'm not sure about some of the changes, please review with care, @caguero :wink:

PR #168, opened 2020-09-01
7 ➡️ 8

chapulina:
#155 was squash-merged, so it looks like this PR is merging a bunch of commits, but it's actually only porting  #164

### DIFF
--- a/log/src/Log.cc
+++ b/log/src/Log.cc
@@ -324,6 +324,14 @@ bool Log::Implementation::InsertMessage(
     const void *_data,
     const std::size_t _len)
 {
+  // \todo Record and playback empty messages. A topic could publish an
+  // Int32 or Int64 message with data=0. In this situation the protobuf
+  // message has size zero. While this is not a problem for protobuf messages,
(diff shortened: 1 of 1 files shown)


chapulina:
* The `ReplayStep` test failing on Windows doesn't seem flaky: https://build.osrfoundation.org/job/ignition_transport-ci-pr_any-windows7-amd64/1220/testReport/junit/(root)/playback/ReplayStep/history/
* `SubThrottled` fails more than succeeds on macOS, we should disable it: https://build.osrfoundation.org/job/ignition_transport-ci-pr_any-homebrew-amd64/902/testReport/(root)/twoProcPubSub/SubThrottled/history/
* `INTEGRATION_playback` is flaky on macOS, we should disable it. This history only shows failures, but note the build numbers: https://build.osrfoundation.org/job/ignition_transport-ci-pr_any-homebrew-amd64/902/testReport/junit/(root)/INTEGRATION_playback/test_ran/history/

caguero (approved):

PR #173, opened 2020-09-04
Make CPPZMQ a PUBLIC dependency

joxoby:
`zmq.hh` is also part of the public API, for example: https://github.com/ignitionrobotics/ign-transport/blob/11c01433aebe1e7377b6e55f49793c226f0bd033/include/ignition/transport/Helpers.hh#L21

Although the fact that `zmq.hh` is included in the API seems like a bug. @caguero WDYT?


caguero (approved):
Thanks! The `cppzmq` header is needed to use the `CPPZMQ_VERSION` macro.

PR #174, opened 2020-09-10
Fix mem leak

JShep1:
Fixing a small memory leak that I came across while profiling ign-gazebo.  Not sure if `sock` needs to be closed before that `return` call though, but since it was being done at the end of the function, I added it before the error-out `return` call as well.


caguero (approved):
Good catch! Looks good to me.

PR #182, opened 2020-09-24
Remove contribute tutorial

caguero:
This is already in the `README.md` file.


chapulina (approved):
LGTM. I think we shouldn't remove it from old collections just in case people are linking to it somewhere.

PR #191, opened 2020-10-22
Test ign-cmake2 revert

mjcarroll:
Test that https://github.com/ignitionrobotics/ign-cmake/pull/120 works correctly

### DIFF
--- a/configure.bat
+++ b/configure.bat
@@ -7,6 +7,7 @@
 call %win_lib% :download_unzip_install libzmq-4.2.3_cppzmq-4.2.2_vc15-x64-dll-MD.zip
 call %win_lib% :download_unzip_install sqlite-3.22.0-vc15-Win64-dll-MD.zip
 call %win_lib% :install_ign_project ign-msgs ign-msgs5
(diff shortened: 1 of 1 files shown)


mjcarroll:
This can be closed once ignitionrobotics/ign-cmake#120 lands.

configure.bat
```diff
@@ -7,6 +7,7 @@
 call %win_lib% :download_unzip_install libzmq-4.2.3_cppzmq-4.2.2_vc15-x64-dll-MD.zip
 call %win_lib% :download_unzip_install sqlite-3.22.0-vc15-Win64-dll-MD.zip
 call %win_lib% :install_ign_project ign-msgs ign-msgs5
+call %win_lib% :install_ign_project ign-cmake revert-115-add_config_search_for_tinyxml2_2
```
chapulina, line 10: I wonder if it makes a difference if it comes before or after `ign-msgs`

chapulina: Interesting, it did both:

`Installing branch ign-cmake2 of ign-cmake`

`Installing branch revert-115-add_config_search_for_tinyxml2_2 of ign-cmake`

But the 2nd one overwrote the first.

It looks like the fix works :+1:

PR #194, opened 2020-11-04
Resolve updated codecheck issues

adlarkin:
This PR is meant to resolve new codecheck issues introduced by https://github.com/ignitionrobotics/ign-cmake/pull/117.

From my understanding, the plan is to forward port the fixes here to all other ignition versions once https://github.com/ignitionrobotics/ign-cmake/pull/117 is merged.

Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

### DIFF
--- a/log/src/cmd/LogCommandAPI.cc
+++ b/log/src/cmd/LogCommandAPI.cc
@@ -15,9 +15,12 @@
  *
 */
 
+#include "LogCommandAPI.hh"
+
 #include <csignal>
 #include <iostream>
 #include <regex>
+#include <string>
 
 #include <ignition/transport/log/Export.hh>
 #include <ignition/transport/log/Playback.hh>
@@ -25,7 +28,6 @@
 #include <ignition/transport/Node.hh>
 #include <ignition/transport/NodeOptions.hh>
 #include "../Console.hh"
-#include "LogCommandAPI.hh"
 
 using namespace ignition;
 ignition::transport::log::PlaybackHandlePtr g_playbackHandler;
--- a/log/src/Descriptor.cc
+++ b/log/src/Descriptor.cc
@@ -15,9 +15,10 @@
  *
 */
 
-#include <string>
-
 #include "Descriptor.hh"
+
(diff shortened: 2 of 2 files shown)


log/src/cmd/LogCommandAPI.cc
```diff
@@ -15,17 +15,19 @@
...
 #include <regex>
+#include <string>
 
 #include <ignition/transport/log/Export.hh>
 #include <ignition/transport/log/Playback.hh>
 #include <ignition/transport/log/Recorder.hh>
 #include <ignition/transport/Node.hh>
 #include <ignition/transport/NodeOptions.hh>
 #include "../Console.hh"
-#include "LogCommandAPI.hh"
```
chapulina, line 28: Just curious, does the new linter require that only the header for this source file be up top?

chapulina (approved):
Just a question about the moved internal headers.

log/src/cmd/LogCommandAPI.cc
adlarkin: It does not - I moved it to follow the [google style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes), which states:

> Include headers in the following order: Related header, C system headers, C++ standard library headers, other libraries' headers, your project's headers.

(note the related header being first) should I leave it the way it was? I made this change (making the related header being the first thing included) in other files I modified for codecheck as well.

chapulina: I think it's fine, it looks like the local codechecker is happy here. It isn't happy in other libraries though, the `tools/code_check.sh` script may be slightly different for those libraries.

adlarkin: I made a [comment](https://github.com/ignitionrobotics/ign-gui/pull/144#discussion_r517073234) about the other library where this is an issue - I think the other library actually had an issue with the QT header, and that there isn't a difference with the `tools/code_check.sh` script.

PR #196, opened 2020-11-06
Add support for bazel buildsystem

mjcarroll:


BUILD.bazel
```diff
@@ -0,0 +1,176 @@
...
+PROJECT_NAME = "ignition-transport"
+
+PROJECT_MAJOR = 9
+
+PROJECT_MINOR = 0
+
+PROJECT_PATCH = 0
+
+# Generates config.hh based on the version numbers in CMake code.
+ign_config_header(
```
mahiuchun, line 36: this rule is not working for me; I got `KeyError: 'Missing definition for PROJECT_NAME'`.

BUILD.bazel
```diff
@@ -0,0 +1,176 @@
...
+        "IGN_TRANSPORT_TEST_DIR='\"./ign_transport/test\"'",
+        "IGN_TEST_LIBRARY_PATH='\"./ign_transport/test\"'",
+    ],
+    deps = [
+        ":ign_transport",
+        "//ign_transport/test:test_utils",
+        "@gtest",
+        "@gtest//:gtest_main",
+    ],
+) for src in test_sources]
```
JShep1, line 176: Think this might need a `exports_files(["src/cmd/cmdtransport.rb.in"])` for `ign_tools`

JShep1 (changes requested):
I did need to add that one line to make it compatible with `ign_tools`

Additionally, is the version of `zmq` brought in by `ign_bazel` correct?  I'm getting the error ```ERROR: /home/john/.cache/bazel/_bazel_john/21b88e8e078a969a58c981cbbc88b951/external/zmq/BUILD.bazel:3:11: @zmq//:zmq: missing input file 'external/zmq/zmq_addon.hpp', owner: '@zmq//:zmq_addon.hpp'

ERROR: /home/john/.cache/bazel/_bazel_john/21b88e8e078a969a58c981cbbc88b951/external/zmq/BUILD.bazel:3:11 1 input file(s) do not exist```

Are you seeing anything like this?

BUILD.bazel
```diff
@@ -0,0 +1,176 @@
+load(
+    "//ign_bazel:build_defs.bzl",
+    "IGNITION_FEATURES",
```
mahiuchun, line 3: Just want to point out `IGNITION_FEATURES` is depending on https://github.com/ignitionrobotics/ign-bazel/pull/28

mjcarroll:
> Additionally, is the version of zmq brought in by ign_bazel correct?

Maybe it's because I'm on focal? I believe that header can be safely removed in this case.

BUILD.bazel
mjcarroll: I'm not sure why that would be the case?  It's the same as the other locations that it is used.  Is you `ign-bazel` at the lastest?

mahiuchun: ign-bazel with https://github.com/ignitionrobotics/ign-bazel/pull/28 works. otherwise I get the above-mentioned error.

BUILD.bazel
mjcarroll: Acknowledged

JShep1 (approved):

JShep1:
LGTM, builds fine

JShep1:
> > Additionally, is the version of zmq brought in by ign_bazel correct?
> 
> Maybe it's because I'm on focal? I believe that header can be safely removed in this case.

Builds fine now on Bionic with all tests green.

log/src/Log_TEST.cc
```diff
@@ -248,6 +248,7 @@ TEST(Log, NullDescriptorUnopenedLog)
 }
 
 //////////////////////////////////////////////////
+/*
 TEST(Log, OpenCorruptDatabase)
```
chapulina, line 252: We shouldn't be disabling tests. Is there a plan to make this test work before merging the PR?

mjcarroll: Yes, it will be restored, I believe I commented this out when it was generally not working on focal.  I will verify.

mahiuchun (approved):

chapulina (changes requested):
Tests need to be uncommented before merging.

chapulina:
I believe this PR is outdated and not being worked on. Please reopen if this starts being worked on again.

Issue #192, opened 2020-10-26
Conformity of current multicast IP


mbmende:
The choice of the multicast IP in https://github.com/ignitionrobotics/ign-transport/blob/1b3a3e0dae80c7b42805d6f54a5a5dfe11f0c27b/include/ignition/transport/Discovery.hh#L1358
should be reconsidered, since it's nonconforming. According to [IANA], "_the range of addresses between 224.0.0.0 and 224.0.0.255, inclusive, is reserved for the use of routing protocols and other low-level topology discovery or maintenance protocols, such as gateway discovery and group membership reporting._" The importance of this range is also the reason, why [RFC4541] requires that "_packets with a destination IP (DIP) address in the 224.0.0.X range which are not IGMP must be forwarded on all ports_" because e.g. routers with missing IGMP join messages could lead to switches splitting up the network. All currently available switches adhere to that recommendation.

The problem with the current choice is that all messages are flooded to the whole network, especially if several instances of gazebo are running with high update rates and >1k packets/s. In our case, with a /22 network and many clients, this leads to a large increase of non directed additional traffic, also to IoT devices and other hosts that don't cope well with massive broadcasts. As this issue might affect a lot of users, even unknowingly, it should be dealt with on a general level instead of in cases where the problem arises.

All other IP ranges defined for multicast are really "multicasted" by switches and hence suited, but since [RFC2365] recommends the 239.255.x.x range for this use, I would suggest to change the hardcoded IP to something like 239.255.0.7 (ROS2 already uses 239.255.0.1 afaik) or make it configurable with a reasonable default value.

[IANA]: https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml#multicast-addresses-13
[RFC4541]: https://tools.ietf.org/html/rfc4541#section-2.1.2
[RFC2365]: https://tools.ietf.org/html/rfc2365#section-6.1

caguero:
Thanks for the detailed explanation. It sounds reasonable to make it configurable with a default option in the range `239.255.x.x`.

caguero:
@mbmende , see pull request #200 . Could you give it a try?

mbmende:
Looks good so far with the latest changes in https://github.com/ignitionrobotics/ign-transport/commit/34f8476bdf1b440a0dccc39bc79d1a5cb6e7be66.

peci1:
[...]

PR #200, opened 2020-11-12
Configurable IP address and port for discovery

caguero:
See issue #192.

How to test it:

Run the `example/publisher` with:

```
IGN_DISCOVERY_MULTICAST_IP=239.255.0.7 IGN_DISCOVERY_MSG_PORT=11000 IGN_DISCOVERY_SRV_PORT=11001 IGN_VERBOSE=1 ./publisher
```

And the `example/subscriber` with:

```
IGN_DISCOVERY_MSG_PORT=11000 IGN_DISCOVERY_SRV_PORT=11001 ./subscriber
```

You should receive the messages as expected. Then, run:

```
./subscriber
```

And you shouldn't get any messages.


tutorials/20_env_variables.md
```diff
@@ -10,6 +10,18 @@ nodes or specify some options when you advertise a topic, it is possible to
...
 Below are descriptions of the available environment variables:
 
+* **IGN_DISCOVERY_MSG_PORT**
+    * *Value allowed*: Any non-negative number in range [0-65535]. In practice
+    you should use the range [1024-65535].
+    * *Description*: UDP port used for message discovery.
+* **IGN_DISCOVERY_MULTICAST_IP**
+    * *Value allowed*: Any multicast IP address
+    * *Description*: Multicast IP address used for communicating all the
+    discovery messages.
```
nkoenig, line 20: Can you add the default IP, or indicate how a user can get access to the default IP?

The same goes for the two new port numbers.

nkoenig (changes requested):

tutorials/20_env_variables.md
caguero: Thanks, added in 34f8476.

nkoenig (approved):

include/ignition/transport/Discovery.hh
```diff
@@ -119,12 +119,15 @@ namespace ignition
       /// \brief Constructor.
       /// \param[in] _pUuid This discovery instance will run inside a
       /// transport process. This parameter is the transport process' UUID.
+      /// \param[in] _ip IP address used for discovery traffic.
       /// \param[in] _port UDP port used for discovery traffic.
       /// \param[in] _verbose true for enabling verbose mode.
       public: Discovery(const std::string &_pUuid,
+                        const std::string &_ip,
```
chapulina, line 126: Is it worth it tick-tocking this? Like adding a new constructor, and using a default value with this contructor?

include/ignition/transport/NodeShared.hh
```diff
@@ -301,11 +301,20 @@ namespace ignition
       /// \brief Mutex to guarantee exclusive access between all threads.
       public: mutable std::recursive_mutex mutex;
```
chapulina, line 303: We should consider PIMPLizing this class

log/src/Recorder.cc
```diff
@@ -188,8 +188,11 @@ Recorder::Implementation::Implementation()
     this->OnMessageReceived(_data, _len, _info);
   };
 
-  this->discovery = std::unique_ptr<MsgDiscovery>(
-        new MsgDiscovery(Uuid().ToString(), NodeShared::kMsgDiscPort));
+  auto shared = NodeShared::Instance();
+
+  this->discovery =
+    std::unique_ptr<MsgDiscovery>(new MsgDiscovery(
```
chapulina, line 194: Could you use `std::make_unique`?

src/NodeShared.cc
```diff
@@ -1818,3 +1796,40 @@ void NodeSharedPrivate::PublishThread()
...
+{
+  int numVal = _defaultValue;
+  std::string strVal;
+  if (env(_envVar, strVal))
+  {
+    try
+    {
+      numVal = std::stoi(strVal);
+    }
+    catch (std::invalid_argument &_e)
```
chapulina, line 1812: `_e` is not used, I think it can be omitted

chapulina (commented):
Just left some high-level comments

include/ignition/transport/Discovery.hh
caguero: We could but honestly I don't think it's worth it because this class is mainly internal.

log/src/Recorder.cc
caguero: > Could you use `std::make_unique`?

Thanks! b8fcb21

src/NodeShared.cc
caguero: > `_e` is not used, I think it can be omitted

Removed in b8fcb21.

include/ignition/transport/NodeShared.hh
caguero: > We should consider PIMPLizing this class

Maybe I can start working on it on a separate branch to see if there are any issues if sounds reasonable.

caguero:
@osrf-jenkins run tests please.

caguero:
@osrf-jenkins run tests please.

include/ignition/transport/Discovery.hh
chapulina: Oh I see. Maybe we can stop installing it in Edifice?

caguero: > Oh I see. Maybe we can stop installing it in Edifice?

Sorry, strictly speaking is not internal because if you include `Node.hh`, it will be recursively included eventually. What I mean is that it's very uncommon that someone will directly use `Discovery.hh` as it's functionality is kind of an internal detail of the library.

chapulina: Ok, got it. Maybe in the future we can move it to a `detail` directory to make this clear to downstream users. I wouldn't be surprised if someone is using this header for something weird, but if we explicitly move it to detail at least we're not committing to tick-tock.

chapulina (approved):
I haven't tried it, but it looks reasonable to me :+1:

PR #204, opened 2020-11-25
Add macOS 10.15 workflow, test <filesystem> support

scpeters:
This adds a macOS 10.15 workflow in order to test support for c++ filesystem on 10.15. I expect it to pass on 10.15 and fail on our jenkins 10.14 builds. Marked as a draft as this is primarily a test to illustrate support on different OS versions.

cc @mjcarroll

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,37 @@
+name: macOS latest
+
+on: [push, pull_request]
+
+jobs:
+  build:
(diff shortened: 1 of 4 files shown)


chapulina:
> Marked as a draft as this is primarily a test to illustrate support on different OS versions.

@scpeters , are there plans to merge this? If the testing and illustration are done, maybe we can close it?

PR #205, opened 2020-12-03
Topic statistics

nkoenig:
My todo list:

- [x]  Do another pass at code cleanup
- [x] Write a tutorial
- [x] Check test coverage.

Requires:

https://github.com/ignitionrobotics/ign-msgs/pull/110

This adds topic statistics which can report:

1. Number of dropped messages.
2. Publication, Age, and Reception statistics

The wire protocol changes when topic statistics are enabled, so both the publisher and subscriber need to set the same environment variable.

Example:

1. `IGN_TRANSPORT_TOPIC_STATISTICS=1 ./examples/build/publisher`
2. `IGN_TRANSPORT_TOPIC_STATISTICS=1 ./examples/build/subscriber_stats`
3. `IGN_TRANSPORT_TOPIC_STATISTICS=1 ign topic -et /statistics`


tutorials/20_env_variables.md
```diff
@@ -64,6 +64,14 @@ Below are descriptions of the available environment variables:
     buffer, so your buffer will grow until you run out of memory (and probably
     crash). If your buffer reaches the maximum capacity data will be dropped.
     * *Default value*: 1000.
+* **IGN_TRANSPORT_TOPIC_STATISTICS**
+    * *Value allowed*: 1/0
+    * *Description*: Enable topic statistics. A value of 1 will eable topic
```
caguero: typo: enable

include/ignition/transport/Discovery.hh
```diff
@@ -1269,7 +1269,10 @@ namespace ignition
       /// \return The discovery version.
       private: uint8_t Version() const
       {
-        return this->kWireVersion;
+        static std::string ignStats;
+        static int topicStats =
+          (env("IGN_TRANSPORT_TOPIC_STATISTICS", ignStats) && ignStats == "1");
```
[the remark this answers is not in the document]
nkoenig: I've multiplied `topicStats` by 100 in a260aa1e8f28c83fc4116b9143f03bd838efdf32. The idea is that this version of ign-transport will be long since EOL'ed when wire transport 110/111 is out.

src/TopicStatistics.cc
```diff
@@ -0,0 +1,269 @@
...
+  stat->set_name("dropped_message_count");
+  stat->set_value(this->dataPtr->droppedMsgCount);
+
+  // Publication statistics
+  msgs::StatisticsGroup *statGroup = _msg.add_statistics_groups();
+  statGroup->set_name("publication_statistics");
+  stat = statGroup->add_statistics();
+  stat->set_type(msgs::Statistic::AVERAGE);
+  stat->set_name("avg_hz");
+  stat->set_value(1.0 / this->dataPtr->publication.Avg());
```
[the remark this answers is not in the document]
nkoenig: Fixed in a260aa1e8f28c83fc4116b9143f03bd838efdf32

caguero (approved):

caguero:
We'll probably need an Ignition Msgs release before merging for not breaking CI.

nkoenig:
@osrf-jenkins run tests please.

src/TopicStatistics.cc
```diff
@@ -0,0 +1,269 @@
...
+  uint64_t now =
+    std::chrono::duration_cast<std::chrono::milliseconds>(
+        std::chrono::steady_clock::now().time_since_epoch()).count();
+
+  if (this->dataPtr->prevPublicationStamp != 0)
+  {
+    this->dataPtr->publication.Update(_stamp -
+        this->dataPtr->prevPublicationStamp);
+    this->dataPtr->reception.Update(now - this->dataPtr->prevReceptionStamp);
+    this->dataPtr->age.Update(now - _stamp);
```
chapulina, line 152: This is triggering several Windows warnings for `uint -> double` conversions:

`warning C4244: 'argument': conversion from 'uint64_t' to 'double', possible loss of data`

The `Update` call takes a double. We either need to explicitly cast or perhaps convert the `uint64_t` stamps to doubles? The documentation says `param[in] _stamp Publication time stamp.`, but I think that tick count may be a confusing stamp to use.

chapulina (commented):
It would be interesting to inspect this through the command line.

---

I increased the required `ign-msgs` version in f4cf7bed1d6bbf9aecad4dfcd569e5ed1c43ff95.

nkoenig:

src/TopicStatistics.cc
chapulina: Thank you for fixing the warnings, @nkoenig . I still have a question whether the number of ticks is a good timestamp though. Wouldn't it be better to use a chrono type, or double for seconds, instead of `uint64`?

nkoenig: I chose this type because it's a well defined compact type and it matches ROS.

chapulina: It's still not clear to me what the stamp is supposed to be. The documentation only says "time stamp", and the implementation seems to be using [milliseconds](https://github.com/ignitionrobotics/ign-transport/pull/205/files#diff-455e7fb8e101230dc5ad17486444880f803615ff54cddfdc4774f03db2fa5b5cR376). I think [ROS 2 uses nanoseconds](https://github.com/ros-tooling/libstatistics_collector/blob/d44a42b1a4746d43073baa17dab7899833ae66f5/include/libstatistics_collector/topic_statistics_collector/received_message_age.hpp#L68).

PR #209, opened 2021-01-04
9 to 10 20210104

nkoenig:
Requires https://github.com/ignitionrobotics/ign-msgs/pull/123


chapulina (changes requested):
There shouldn't be an `ign-transport10` branch yet. That's the current version under development. Version 10 is currently on `main` (which used to be `default` on BitBucket)

caguero:
There's a conflict in `src/NodeShared.cc`.

caguero:
> There's a conflict in `src/NodeShared.cc`.

Fixed in 2273d63.

chapulina (changes requested):
There seems to be 2 things combined in this PR. The merge from 9, and the bump of `ign-msgs` to version 7. I think if it was just the merge, it would probably be good to merge. But the bump in dependency still needs updates on CI.

nkoenig:
> There seems to be 2 things combined in this PR. The merge from 9, and the bump of `ign-msgs` to version 7. I think if it was just the merge, it would probably be good to merge. But the bump in dependency still needs updates on CI.

Right, we need to make an ign-msgs7 prerelease.

chapulina:
homebrew CI needs https://github.com/osrf/homebrew-simulation/pull/1293

chapulina (approved):

PR #238, opened 2021-04-13
Bumps in fortress : ign-transport11

chapulina:
See https://github.com/ignition-tooling/release-tools/issues/428

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -1,6 +1,6 @@
 libignition-cmake2-dev
 libignition-math6-dev
-libignition-msgs7-dev
+libignition-msgs8-dev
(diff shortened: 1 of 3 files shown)


adlarkin (approved):

adlarkin (commented):
Not sure if it matters, but it looks like the `bitbucket-pipelines.yml` is out of date: https://github.com/ignitionrobotics/ign-transport/blob/bump_fortress_ign-transport11/bitbucket-pipelines.yml#L30

chapulina:
> Not sure if it matters, but it looks like the bitbucket-pipelines.yml is out of date:

I haven't been worrying about that too much since only the first line is still used.

PR #240, opened 2021-04-23
Remove unnecessary copy and assignment operators

mjcarroll:
Targeting `main`, because I'm not sure if this breaks API.

We were defining copy and assignment operators that aren't necessary and broke rule of three/rule of five.

### DIFF
--- a/include/ignition/transport/Publisher.hh
+++ b/include/ignition/transport/Publisher.hh
@@ -58,10 +58,6 @@ namespace ignition
                         const std::string &_nUuid,
                         const AdvertiseOptions &_opts);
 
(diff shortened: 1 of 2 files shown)


caguero (commented):
Do we need to deprecate them before removing them first?

mjcarroll:
> Do we need to deprecate them before removing them first?

Because the compiler will implicitly generate these functions if they aren't explicitly defined, we aren't really "removing" them per se, so I don't believe a deprecation cycle is necessary.

caguero (approved):
Sounds good to me, then!

chapulina (approved):

PR #241, opened 2021-04-23
Remove unnecessary copy and assignment operators

mjcarroll:
#240 but for ign-transport10 to see if it breaks ABI.

### DIFF
--- a/include/ignition/transport/Publisher.hh
+++ b/include/ignition/transport/Publisher.hh
@@ -58,10 +58,6 @@ namespace ignition
                         const std::string &_nUuid,
                         const AdvertiseOptions &_opts);
 
-      /// \brief Copy constructor.
-      /// \param[in] _other Other Publisher object.
-      public: Publisher(const Publisher &_other);
-
       /// \brief Destructor.
       public: virtual ~Publisher() = default;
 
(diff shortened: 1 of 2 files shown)


scpeters:
> to see if it breaks ABI.

seems like it does

chapulina:
:confused: 

```
Removed Symbols  4 
Publisher.hh
namespace ignition::transport::v10
MessagePublisher::__ct [C1] ( MessagePublisher const& _other )
_ZN8ignition9transport3v1016MessagePublisherC1ERKS2_

MessagePublisher::__ct [C2] ( MessagePublisher const& _other )
_ZN8ignition9transport3v1016MessagePublisherC2ERKS2_

ServicePublisher::__ct [C1] ( ServicePublisher const& _other )
_ZN8ignition9transport3v1016ServicePublisherC1ERKS2_

ServicePublisher::__ct [C2] ( ServicePublisher const& _other )
_ZN8ignition9transport3v1016ServicePublisherC2ERKS2_
```

Note it didn't complain about the `Publisher` though.

mjcarroll:
Trying an alternative approach here.  This suppresses the warnings and also removes the implementation in favor of the `default` impl.  This shouldn't break ABI afaict.

scpeters:
nice work, ABI checker is clean

scpeters (approved):
approved, pending unfinished CI builds

PR #245, opened 2021-05-12
Consider all network interfaces when checking HOST option.

caguero:
Signed-off-by: Carlos Agüero <caguero@openrobotics.org>

# 🦟 Bug fix

Fixes the following issue:

One publisher sets `IGN_IP` to `127.0.0.1` and advertises a topic with `Scope_T::Host`. With these settings, a default subscriber is unable to receive any messages.

## Summary

The problem was that we were only checking if the source IP received was matching the host address used by the receiver. Now, we're checking if the source IP received matches any of the network interfaces or localhost.

To test (from the `examples` directory:

Terminal 1: `IGN_IP=127.0.0.1 ./publisher`
Terminal 2: `./subscriber`

## Checklist
- [ ] Added tests
- [X] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

### DIFF
--- a/include/ignition/transport/Discovery.hh
+++ b/include/ignition/transport/Discovery.hh
@@ -898,6 +898,10 @@ namespace ignition
           this->SendUnicast(msg);
         }
 
+        bool isSenderLocal = (std::find(this->hostInterfaces.begin(),
+          this->hostInterfaces.end(), _fromIp) != this->hostInterfaces.end()) ||
+          _fromIp == "127.0.0.1";
+
         // Update timestamp and cache the callbacks.
         DiscoveryCallback<Pub> connectCb;
         DiscoveryCallback<Pub> disconnectCb;
@@ -923,7 +927,7 @@ namespace ignition
             // Check scope of the topic.
             if ((publisher.Options().Scope() == Scope_t::PROCESS) ||
                 (publisher.Options().Scope() == Scope_t::HOST &&
-                 _fromIp != this->hostAddr))
+                 !isSenderLocal))
             {
               return;
             }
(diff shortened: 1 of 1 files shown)


peci1:
Thanks, it works. I tested both the simple example and a more complex one with the comms model visualization service in SubT.

I have just one comment to the implementation: `127.0.0.1` is not the only localhost address. It's actually `127.255.255.255/8` that's reserved for loopback devices.

caguero:
> Thanks, it works. I tested both the simple example and a more complex one with the comms model visualization service in SubT.
> 
> I have just one comment to the implementation: `127.0.0.1` is not the only localhost address. It's actually `127.255.255.255/8` that's reserved for loopback devices.

What do you suggest? Anything that starts with `127.` should be considered localhost?

peci1:
Yes, that's my suggestions.

caguero:
> Yes, that's my suggestion.

Sure, updated in fc1c7da.

adlarkin (approved):

include/ignition/transport/Discovery.hh
```diff
@@ -898,6 +898,10 @@ namespace ignition
           this->SendUnicast(msg);
         }
 
+        bool isSenderLocal = (std::find(this->hostInterfaces.begin(),
+          this->hostInterfaces.end(), _fromIp) != this->hostInterfaces.end()) ||
+          (_fromIp.find("127.") == 0);
```
mjcarroll, line 903: Is there any need to make this also work for ipv6?

peci1: not from my view, but generally it would be good

caguero: Probably out of scope for now.

PR #279, opened 2021-11-25
Use libexec to install lib binaries

j-rivero:
We are installing the cli11 binaries in under the standard library path which is something that does not fit well according to the [FHR](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrlibexec). libexec seems a better place for them:

```
4.7.1. Purpose

/usr/libexec includes internal binaries that are not intended to be executed directly by users or shell scripts. Applications may use a single subdirectory under /usr/libexec.
```

### DIFF
--- a/src/cmd/CMakeLists.txt
+++ b/src/cmd/CMakeLists.txt
@@ -22,7 +22,7 @@ target_link_libraries(${topic_executable}
   ignition-utils${IGN_UTILS_VER}::cli
   ${PROJECT_LIBRARY_TARGET_NAME}
 )
-install(TARGETS ${topic_executable} DESTINATION ${IGN_LIB_INSTALL_DIR}/ignition/${IGN_DESIGNATION}${PROJECT_VERSION_MAJOR}/)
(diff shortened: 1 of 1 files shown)


caguero (approved):

chapulina:
Do we need any changes to the `-release` repo?

j-rivero:
> Do we need any changes to the `-release` repo?

Yes! https://github.com/ignition-release/ign-transport11-release/pull/4

mjcarroll:
The CLI binaries can still be executed standalone by a user if they don't wish to use `ign`, so I don't know if `libexec` is the correct location for them.

chapulina:
@mjcarroll has a good point. Maybe we need to create new `ignition-transport11` packages for the binaries?

chapulina:
> Maybe we need to create new ignition-transport11 packages for the binaries?

I believe this is what's being suggested for `plugin2` here, right @j-rivero ?

* https://github.com/ignition-release/ign-plugin2-release/pull/5

j-rivero:
> 

Yep, the same.

mjcarroll (approved):

j-rivero:
Updated the -release repo PR https://github.com/ignition-release/ign-transport11-release/pull/4

PR #296, opened 2022-03-10
Use new ignition-specific formatter

mjcarroll:
Requires ignitionrobotics/ign-utils#42


chapulina (commented):
Can we use it on `service_main` too?

https://github.com/gazebosim/gz-transport/blob/ign-transport11/src/cmd/service_main.cc

mjcarroll:
> Can we use it on service_main too?

Sure thing, I'm not sure why that wasn't included originally, maybe this was just a prototype.

caguero (commented):
I made very minor tweaks in [fd9f63b](https://github.com/gazebosim/gz-transport/pull/296/commits/fd9f63b56e7ad99cd0fb879154a1aba07be29bf9).

caguero (approved):

mjcarroll:
I actually meant to add `log` to this as well, but that can probably be done in a follow-up.

PR #313, opened 2022-05-24
Suppress a new Jammy warning

mjcarroll:
`recvfrom` returns an `ssize_t`, which is signed.

```
/home/mjcarroll/workspaces/gz_garden/src/gz-transport/include/gz/transport/Discovery.hh:840:27: warning: comparison is always false due to limited range of data type [-Wtype-limits]
  840 |         else if (received < 0)
```


ahcorde (approved):

mjcarroll:
Windows doesn't have `ssize_t`, this will take another look.

PR #316, opened 2022-05-31
Update GoogleTest to latest version

mjcarroll:
Import latest version of GoogleTest from upstream and adjust build accordingly.


caguero (commented):
CI seems unhappy.

PR #335, opened 2022-07-22
Add tracing to the discovery test

mjcarroll:
Adds a whole bunch of output to figure out why the discovery unit test is failing all of a sudden.

PR #336, opened 2022-07-26
Remove problematic discovery test

mjcarroll:
This test has recently appeared to fail, we have some on-going diagnosis in #335.  In the meantime, I think it makes sense to disable to get CI back into a reasonable state.

### DIFF
--- a/src/Discovery_TEST.cc
+++ b/src/Discovery_TEST.cc
@@ -29,6 +29,19 @@
 #include "gz/transport/Uuid.hh"
 #include "test_config.hh"
 
+// Temporarily introduce a "DISABLED_ON_LINUX" macro.
+// It currently does not exist upstream.
+// This can be removed when it is in upstream gz-utils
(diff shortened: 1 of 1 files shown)


chapulina:
Can we just disable the test on the platforms where it fails, instead of removing it?

mjcarroll:
> Can we just disable the test on the platforms where it fails, instead of removing it?

We can, as it turns out, we don't have a macro to disable a test only on Linux, I suppose it's never been needed before?

chapulina (approved):

mjcarroll:
New warnings on windows are to be addressed elsewhere.

chapulina:
Would you be able to backport this to Citadel and merge forward to Fortress? Thanks!

mjcarroll:
> Would you be able to backport this to Citadel and merge forward to Fortress?

Done: https://github.com/gazebosim/gz-transport/pull/339

PR #351, opened 2022-09-02
Added Node::RequestRaw

nkoenig:

# 🎉 New feature

## Summary

Adds `Node::RequestRaw`, which is similar to `PublishRaw`. In this case, raw serialized request and response strings are used instead of protobuf messages.

Needed by https://github.com/gazebosim/gz-launch/pull/187

## Test it

Run the `requester_raw` example program.

## Checklist
- [x] Added tests
- [x] Added example and/or tutorial

### DIFF
--- a/example/requester_raw.cc
+++ b/example/requester_raw.cc
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#include <iostream>
+#include <ignition/msgs.hh>
+#include <ignition/transport.hh>
+
+//////////////////////////////////////////////////
+int main(int argc, char **argv)
+{
+  // Create a transport node.
+  ignition::transport::Node node;
+
+  // Prepare the input parameters.
+  ignition::msgs::StringMsg req;
+  req.set_data("HELLO");
+
+  bool result;
+  unsigned int timeout = 5000;
+
+  std::string reqStr, repStr, repMsgType;
+  req.SerializeToString(&reqStr);
+
+  // Request the "/echo" service.
+  bool executed = node.RequestRaw("/echo", reqStr, "ignition.msgs.StringMsg",
+      timeout, repStr, repMsgType, result);
+
+  if (executed)
+  {
+    if (result)
+    {
+      if (repMsgType == "ignition.msgs.StringMsg")
+      {
+        ignition::msgs::StringMsg rep;
+        rep.ParseFromString(repStr);
+        std::cout << "Response: [" << rep.data() << "]" << std::endl;
+      }
+      else
+      {
+        std::cout << "Expected reply message type of ignition.msgs.StringMsg "
+          << "received " << repMsgType << std::endl;
+      }
+    }
+    else
+      std::cout << "Service call failed" << std::endl;
+  }
+  else
+    std::cerr << "Service call timed out" << std::endl;
+}
--- a/src/Node.cc
+++ b/src/Node.cc
@@ -1036,3 +1036,33 @@ bool Node::SubscribeHelper(const std::string &_fullyQualifiedTopic)
 {
   return this->dataPtr->SubscribeHelper(_fullyQualifiedTopic);
 }
+
+/////////////////////////////////////////////////
+bool Node::RequestRaw(const std::string &_topic,
+    const std::string &_request, const std::string &_requestType,
+    unsigned int _timeout,
+    std::string &_response, std::string &_responseType,
+    bool &_result)
+{
+  std::vector<ServicePublisher> publishers;
+  this->ServiceInfo(_topic, publishers);
+
+  if (publishers.empty())
+  {
+    std::cerr << "Node::RequestRaw(): Error getting response type for "
+      << "service [" << _topic << "]\n";
+    return false;
+  }
+  _responseType =publishers.front().RepTypeName();
+
+  std::unique_ptr<google::protobuf::Message> req =
+    msgs::Factory::New(_requestType);
+  req->ParseFromString(_request);
+
+  std::unique_ptr<google::protobuf::Message> res =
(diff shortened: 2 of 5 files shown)


src/Node.cc
```diff
@@ -1036,3 +1036,33 @@ bool Node::SubscribeHelper(const std::string &_fullyQualifiedTopic)
...
+
+/////////////////////////////////////////////////
+bool Node::RequestRaw(const std::string &_topic,
+    const std::string &_request, const std::string &_requestType,
+    unsigned int _timeout,
+    std::string &_response, std::string &_responseType,
+    bool &_result)
+{
+  std::vector<ServicePublisher> publishers;
+  this->ServiceInfo(_topic, publishers);
```
caguero, line 1048: I think this approach will work when the responser is ready before the requester_raw tries to request. However if you launch the responser after the requester it won't work. I'd remove all the code from line `1047` to `1057` (`Request()` already handles this for us).

src/Node.cc
```diff
@@ -1036,3 +1036,33 @@ bool Node::SubscribeHelper(const std::string &_fullyQualifiedTopic)
...
+  if (publishers.empty())
+  {
+    std::cerr << "Node::RequestRaw(): Error getting response type for "
+      << "service [" << _topic << "]\n";
+    return false;
+  }
+  _responseType = publishers.front().RepTypeName();
+
+  std::unique_ptr<google::protobuf::Message> req =
+    msgs::Factory::New(_requestType);
```
caguero, line 1059: You can add the data directly to `New()` with `msgs::Factory::New(_requestType, _request);`

src/Node.cc
```diff
@@ -1036,3 +1036,33 @@ bool Node::SubscribeHelper(const std::string &_fullyQualifiedTopic)
...
+  req->ParseFromString(_request);
+
```
caguero, line 1061: Can we check if `New()` succeeded? Something like:

```
if (!req)
{
  std::cerr << "Unable to create request of type[" << _requestType << "] "
                << "with data[" << _request << "].\n";
  return false;
}
```

src/Node.cc
```diff
@@ -1036,3 +1036,33 @@ bool Node::SubscribeHelper(const std::string &_fullyQualifiedTopic)
...
+  std::unique_ptr<google::protobuf::Message> res =
```
caguero, line 1062: Same here, can we check if `res` succeeded?

caguero (changes requested):
Looking good, just minor suggestions.

src/Node.cc
nkoenig: This is here so that I can get the responder's message type. This function doesn't require the user to specify the response type. Instead it gets the response type from the responder.

src/Node.cc
nkoenig: That doesn't work in this case. The `_request` is a serialized message. The `msgs::Factory::New(_requestType, _request);` expects an unserialized protobuf message.

src/Node.cc
nkoenig: Added in [5c404e9](https://github.com/gazebosim/gz-transport/pull/351/commits/5c404e9753c65829e42c2ed506a6fe74ef6ae81b)

src/Node.cc
nkoenig: Added in [5c404e9](https://github.com/gazebosim/gz-transport/pull/351/commits/5c404e9753c65829e42c2ed506a6fe74ef6ae81b)

src/Node.cc
caguero: Oh I see, I didn't realize that `_responseType` wasn't an input parameter. I see two problems with the current code:

1. It's not enough to uniquely identify a service with just a topic name and the request type. It's possible and legal that two nodes advertise services with the same topic name, same request type but different response type. In that case, `Request()` selects the right available service. Do you have a particular use case for not passing the response type?

2. `ServiceInfo()` checks if we have certain information right now. This is fragile as it requires that the service provider was up and running before checking this function. For overcoming this issue we might need to reuse the existing `Request()` if possible or implement it in a similar way storing a request handler.

nkoenig: My use case is through websocket connections, which can't specify the response type using the existing protocol.

In my experience most/all services are unique based solely on their name. I could add a note to the documentation of this function about it's behavior.

nkoenig: @caguero I refactored this function so that it accepts the `_responseType` instead of discovering in. See 
f1d44d2.

src/Node.cc
```diff
@@ -1036,3 +1036,32 @@ bool Node::SubscribeHelper(const std::string &_fullyQualifiedTopic)
...
+  if (!req)
+  {
+    std::cerr << "Unable to create request of type[" << _requestType << "].\n";
+    return false;
+  }
+  req->ParseFromString(_request);
+
+  std::unique_ptr<google::protobuf::Message> res =
+    msgs::Factory::New(_responseType);
+  if (!req)
```
caguero, line 1057: I think this should be `res`.

example/requester_raw.cc
```diff
@@ -0,0 +1,63 @@
...
+  req.set_data("HELLO");
+
+  bool result;
+  unsigned int timeout = 5000;
+
+  std::string reqStr, repStr, repMsgType;
+  req.SerializeToString(&reqStr);
+
+  // Request the "/echo" service.
+  bool executed = node.RequestRaw("/echo", reqStr, "ignition.msgs.StringMsg",
```
caguero, line 39: I think this call needs to be updated: the order of the parameters doesn't match the signature and `repMsgType` should be initialized.

src/Node.cc
```diff
@@ -1036,3 +1036,32 @@ bool Node::SubscribeHelper(const std::string &_fullyQualifiedTopic)
...
+  {
+    std::cerr << "Unable to create response of type["
+      << _responseType << "].\n";
+    return false;
+  }
+
+  bool executed = this->Request(_topic, *req, _timeout, *res, _result);
+  res->SerializeToString(&_response);
```
caguero, line 1065: Nit: Check the result of `SerializeToString()`. Another idea is to return `executed && res->SerializeTostring()`.

caguero (changes requested):

src/Node.cc
nkoenig: Fixed in [3036cb7](https://github.com/gazebosim/gz-transport/pull/351/commits/3036cb7515c57e2adece1f6bf56fb455d2b26bf5)

example/requester_raw.cc
nkoenig: Fixed in [3036cb7](https://github.com/gazebosim/gz-transport/pull/351/commits/3036cb7515c57e2adece1f6bf56fb455d2b26bf5)

src/Node.cc
nkoenig: Fixed in [3036cb7](https://github.com/gazebosim/gz-transport/pull/351/commits/3036cb7515c57e2adece1f6bf56fb455d2b26bf5)

caguero (approved):

PR #372, opened 2022-11-30
Fix CLI configuration install path to ignition

mjcarroll:
# 🦟 Bug fix

## Summary

When `fortress` and `garden` are installed on the same system, there is a case where `ign topic` doesn't appear in the list of available commands for `fortress`, and `garden`'s `gz` command will not work with the error:

```
Configuration error: File [/usr/share/gz//transport12.yaml] contains a command already registered by ignition-transport.
```

It appears that somewhere in the forward porting of the renaming, we started erroneously installing `transport11.yaml` into the `/usr/share/gz` directory.  Specifically, the issue was introduced here (thanks @scpeters) https://github.com/gazebosim/gz-transport/commit/e31248efa946dab6efb83d674b1aac094955bdf7

This restores the installation location to the correct place.

## Checklist
- [ ] Added tests

### DIFF
--- a/conf/CMakeLists.txt
+++ b/conf/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Used only for internal testing.
-set(gz_library_path "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/ignition/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
+set(gz_library_path "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/gz/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
 
 # Generate a configuration file for internal testing.
 # Note that the major version of the library is included in the name.
@@ -13,7 +13,7 @@ file(GENERATE
   INPUT "${CMAKE_CURRENT_BINARY_DIR}/${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}.yaml.configured")
 
 # Used for the installed version.
-set(gz_library_path "${CMAKE_INSTALL_PREFIX}/lib/ruby/ignition/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
+set(gz_library_path "${CMAKE_INSTALL_PREFIX}/lib/ruby/gz/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
 
 # Generate the configuration file that is installed.
 # Note that the major version of the library is included in the name.
--- a/conf/CMakeLists.txt
+++ b/conf/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Used only for internal testing.
-set(gz_library_path "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/gz/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
+set(gz_library_path "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/ignition/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
 
 # Generate a configuration file for internal testing.
 # Note that the major version of the library is included in the name.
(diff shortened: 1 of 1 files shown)


mjcarroll:
FYI @andermi

nkoenig (approved):

conf/CMakeLists.txt
```diff
@@ -24,4 +24,4 @@ configure_file(
 
 # Install the yaml configuration files in an unversioned location.
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}.yaml
-  DESTINATION ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATAROOTDIR}/gz/)
+  DESTINATION ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATAROOTDIR}/ignition/)
```
scpeters, line 27: this is the line that most needs to change

conf/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ file(GENERATE
   INPUT "${CMAKE_CURRENT_BINARY_DIR}/${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}.yaml.configured")
 
 # Used for the installed version.
-set(gz_library_path "${CMAKE_INSTALL_PREFIX}/lib/ruby/gz/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
+set(gz_library_path "${CMAKE_INSTALL_PREFIX}/lib/ruby/ignition/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
```
scpeters, line 16: this install location needs to be consistent with the value in `src/cmd/CMakeLists.txt`:

* https://github.com/gazebosim/gz-transport/blob/ign-transport11/src/cmd/CMakeLists.txt#L109

so I think you could either revert this line, or change `src/cmd/CMakeLists.txt` as well

conf/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # Used only for internal testing.
-set(gz_library_path "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/gz/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
+set(gz_library_path "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/ignition/cmd${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}")
```
scpeters, line 2: this path needs to be consistent with the value in `src/cmd/CMakeLists.txt`:

* https://github.com/gazebosim/gz-transport/blob/ign-transport11/src/cmd/CMakeLists.txt#L68

so you could either revert this line or change `src/cmd/CMakeLists.txt` as well

conf/CMakeLists.txt
mjcarroll: In order to be correct, this needs to be in `<prefix>/lib/ruby/ignition` though, correct?

scpeters: there are two files `*.yaml` and `cmd*.rb`:

* the `*.yaml` file is the entry point. It absolutely needs to be in `share/ignition` on Fortress. This file has the path to the `cmd*.rb` embedded within. The location of the `cmd*.rb` doesn't matter so much as long as the `*.yaml` file has the correct path to the `cmd*.rb`
* the `cmd*.rb` file has been installed in `<prefix>/lib/ruby/ignition` for citadel and fortress (until https://github.com/gazebosim/gz-transport/pull/368), and in `<prefix>/lib/ruby/gz` on garden and recently on fortress. The actual install location of `cmd*.rb` doesn't matter so much as long as it matches where the `*.yaml` file thinks it is. The CI is currently failing on this branch because we have changed where the `*.yaml` file thinks the `cmd*.rb` is installed (the `gz_library_path ` variable above) without actually changing the install command in [src/cmd/CMakeLists.txt:109](https://github.com/gazebosim/gz-transport/blob/ign-transport11/src/cmd/CMakeLists.txt#L109). So you can choose to revert these changes to `gz_library_path` or to update `src/cmd/CMakeLists.txt` to keep them consistent

mjcarroll: Okay, after thinking about this for a bit, I opted to keep the rest of the `gz` changes and only change the `yaml` file installation location.  While this introduces an internal inconsistency in the transport package, I think that it makes more sense to rename as many things to `gz` as possible without too many iterations.  Thoughts?

scpeters (approved):

conf/CMakeLists.txt
scpeters: make sense to me!

Issue #380, opened 2023-01-23
Callbacks can be fired after topic unsubscription


mjcarroll:
## Environment
* OS Version: Ubuntu 22.04
* Source or binary build?  `garden` sources

## Description
* Expected behavior:

After unsubscription, the callbacks associated with a topic should no longer be fired.
This is especially important if there is associated state with the callbacks (eg a lambda) that could cause nullptr dereferencing.

* Actual behavior: The callbacks continue to fire (at least once)

## Steps to reproduce

I have added a test here: https://github.com/gazebosim/gz-transport/blob/mjcarroll/callback_scope_test/test/integration/callback_scope_TEST.cc

If working correctly, the callback function should not be called after the unsubscribe.  In reality, the `FAIL()` condition is reached.

## Output

```
./bin/INTEGRATION_callback_scope_TEST 
Running main() from /home/mjcarroll/workspaces/gz_garden/src/gz-transport/test/gtest_vendor/src/gtest_main.cc
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from CallbackScope
[ RUN      ] CallbackScope.CleanupCorrectly
/home/mjcarroll/workspaces/gz_garden/src/gz-transport/test/integration/callback_scope_TEST.cc:39: Failure
Failed
[  FAILED  ] CallbackScope.CleanupCorrectly (10 ms)
[----------] 1 test from CallbackScope (10 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (10 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] CallbackScope.CleanupCorrectly
```

PR #381, opened 2023-02-07
Prevent invoking callbacks after a node unsubscribes to a topic

iche033:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-transport/issues/380

## Summary

This PR builds on top of the integration test in the `mjcarroll/callback_scope_test` branch and the test should now pass with these changes.

The fix is to loop through the publish queue and remove any outdated handlers that's responsible for invoking callbacks for the topic that's already been unsubscribed.

One important change that I need feedback on is the change of `pubQueue` from `std::queue` to `std::list`. This is done so that I can loop through the STL data structure

## Checklist
- [x] Added tests

### DIFF
--- a/include/gz/transport/Node.hh
+++ b/include/gz/transport/Node.hh
@@ -753,12 +753,6 @@ namespace gz
       /// \return True on success.
       private: bool SubscribeHelper(const std::string &_fullyQualifiedTopic);
 
-      /// \brief Helper function to remove handlers from the shared publish
-      /// queue. This is called when the node unsubscribes to a topic
-      /// \param[in] _topic Topic that the node unsubcribed to.
-      /// \return True on success.
-      private: bool RemoveHandlersFromPubQueue(const std::string &_topic);
-
 #ifdef _WIN32
 // Disable warning C4251 which is triggered by
 // std::unique_ptr
--- a/include/gz/transport/Node.hh
+++ b/include/gz/transport/Node.hh
@@ -753,8 +753,8 @@ namespace gz
       /// \return True on success.
       private: bool SubscribeHelper(const std::string &_fullyQualifiedTopic);
 
-      /// \brief Helper function to remove handlers from the shared publish queue.
-      /// This is called when the node unsubscribes to a topic
+      /// \brief Helper function to remove handlers from the shared publish
+      /// queue. This is called when the node unsubscribes to a topic
       /// \param[in] _topic Topic that the node unsubcribed to.
       /// \return True on success.
       private: bool RemoveHandlersFromPubQueue(const std::string &_topic);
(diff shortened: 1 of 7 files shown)


include/gz/transport/Node.hh
```diff
@@ -753,6 +753,12 @@ namespace gz
       /// \return True on success.
       private: bool SubscribeHelper(const std::string &_fullyQualifiedTopic);
 
+      /// \brief Helper function to remove handlers from the shared publish
+      /// queue. This is called when the node unsubscribes to a topic
+      /// \param[in] _topic Topic that the node unsubcribed to.
+      /// \return True on success.
+      private: bool RemoveHandlersFromPubQueue(const std::string &_topic);
```
ahcorde, line 760: The return value is not used.

ahcorde (dismissed):

include/gz/transport/Node.hh
iche033: added check for return value in [814ca8d](https://github.com/gazebosim/gz-transport/pull/381/commits/814ca8ddd4fa783ed7788c5c0072b87c80ae3c36)

caguero, line 760: Move this function to `NodePrivate.hh` perhaps?

src/Node.cc
```diff
@@ -607,6 +608,15 @@ bool Node::Unsubscribe(const std::string &_topic)
   this->dataPtr->shared->localSubscribers.RemoveHandlersForNode(
         fullyQualifiedTopic, this->dataPtr->nUuid);
 
+  // Remove handlers from shared pubQueue to avoid invoking callbacks after
+  // unsuscribing to the topic
+  if (!this->RemoveHandlersFromPubQueue(topic))
```
caguero, line 613: Consider to call this function right before line `605` (before taking the `shared->mutex`):
  * The faster we remove the handlers, the less chance they'll get used.
  * We avoid taking two consecutive mutexes, which always makes me think on a potential deadlock.

caguero (commented):
It looks good to me. Just two minor comments to consider.

include/gz/transport/Node.hh
iche033: moved. [29eb916](https://github.com/gazebosim/gz-transport/pull/381/commits/29eb9166874a139e7cd478f7683c6c4091221005)

src/Node.cc
iche033: sounds good, moved call earlier in function. [29eb916](https://github.com/gazebosim/gz-transport/pull/381/commits/29eb9166874a139e7cd478f7683c6c4091221005)

mjcarroll:
I think this approach is good.  For completeness, can you add a test that covers local and raw subscribers?

iche033:
ok just updated test to include raw subscribers. [9945415](https://github.com/gazebosim/gz-transport/pull/381/commits/99454158d2973ba5eacbf9ab932b847a92692c80)

caguero (approved):

mjcarroll (approved):

PR #382, opened 2023-02-10
Pass std::function by value to Node::Subscribe

nkoenig:

# 🎉 New feature

Yes this does break API in a released version of transport. However, I believe it doesn't functionally change how `Node::Subscribe` operates, and should not affect downstream users.

The [Node::Subscribe](https://github.com/gazebosim/gz-transport/blob/ign-transport8/include/gz/transport/detail/Node.hh#L56) functions that take a `std::function` use pass by non-const reference. The provided `_cb` function is then copy-captured in a lambda function, or copied in [SubscriptionHandler::SetCallback](https://github.com/gazebosim/gz-transport/blob/ign-transport8/include/gz/transport/SubscriptionHandler.hh#L193).

Passing by value, which this PR proposes, has the same result as passing by reference. A copy is ultimately made.

You can read about [lambda captures here](https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture).

You can look at what the compiler does [here](https://godbolt.org/z/ac45KfxMo). Change the parameter types and you should see the same set of operations.

Given that, I'd like to make an exception to our API rule. As a side benefit, users will gain the ability to do the following:

```
    n.Subscribe([](const Foo& msg){ return; });
```

Thanks to @mjcarroll for the help and @codebot for the suggestion.

## Checklist
- [ ] Added tests
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

### DIFF
--- a/include/gz/transport/Node.hh
+++ b/include/gz/transport/Node.hh
@@ -273,7 +273,7 @@ namespace ignition
       public: template<typename MessageT>
       bool Subscribe(
           const std::string &_topic,
-          std::function<void(const MessageT &_msg)> &_callback,
+          std::function<void(const MessageT &_msg)> _callback,
           const SubscribeOptions &_opts = SubscribeOptions());
 
       /// \brief Subscribe to a topic registering a callback.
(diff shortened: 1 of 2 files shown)


caguero (approved):
It looks reasonable to me. ABI seems safe too.

mjcarroll (approved):

PR #411, opened 2023-06-20
Python Bindings for Publisher, Subscriber and Service Request features.

### LINKED ISSUE #387: Add Python bindings
## Desired behavior

Add Python bindings to `gz-transport` (and `gz-msgs`).

## Alternatives considered

- Use `ros_gz` bridge and the Python support in ROS / ROS2.

## Implementation suggestion

Upstream the bindings for `gz-msgs` and `gz-transport` from `https://github.com/srmainwaring/gz-python`.

## Additional context

From time to time there are requests for Python bindings to control the simulation. A number of the requirements can be met by using the messaging infrastructure to control and interact with the simulation.

- https://community.gazebosim.org/t/python-bindings-for-gazebo-ignition/360
- https://github.com/gazebosim/gz-sim/issues/1905
- https://github.com/gazebosim/gz-transport/pull/386

Voldivh:
# 🎉 New feature

Closes #387 

## Summary

This PR creates the python bindings for the publisher, subscriber and request features from the repository. In order to avoid creating a new dependency to be able to create python bindings from protobuf messages, it was decided to use the methods that publishes, subscribes and make services request with serialized messages, i.e, use the methods `PublishRaw`, `SubscribeRaw` and `RequestRaw`. The bindings for these methods are wrapped by python methods  in order to give the user an API similar to the one we have in C++.

We would like to acknowledge and give credit to @srmainwaring for his [code](https://github.com/srmainwaring/gz-python) which served as a reference and inspiration for this implementation. The main difference with this implementation is that we are using the methods that uses serialized messages instead of creating bindings with the `pybind11_protobuf` library to the non-serialized methods.

## Test it

There are some [examples](https://github.com/Voldivh/gz-transport/tree/voldivh/pybind11_bindings/python/examples) that were created as smoke test in order to test out the functionality. In order to use them you would have to:

1. Compile from source this branch
2. Update the `PYTHONPATH` variable:
```bash
export PYTHONPATH=$PYTHONPATH:<path_to_ws>/install/lib/python
```
3. Run the python script for the example you would like to try out:
```bash
python3 <path_to_ws>/src/gz-transport/python/examples/<example_to_run>
```
**Note**: The service requester example will only work if there is a server running that listens to `/echo` topic and expects a `gz.messages.StringMsg` as the request. One way to do so is to follow the instructions on the [examples folder](https://github.com/Voldivh/gz-transport/blob/voldivh/pybind11_bindings/example/README.md) and run the `responser` executable.


python/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+    pubSub_TEST
+  )
+  execute_process(COMMAND "${Python3_EXECUTABLE}" -m pytest --version
+    OUTPUT_VARIABLE PYTEST_output
+    ERROR_VARIABLE  PYTEST_error
+    RESULT_VARIABLE PYTEST_result)
+  if(${PYTEST_result} EQUAL 0)
+    set(pytest_FOUND TRUE)
+  else()
+    message("")
```
ahcorde, line 80: remove

CMakeLists.txt
```diff
@@ -117,14 +132,45 @@ gz_find_package(SQLite3
...
+    else()
+      GZ_BUILD_WARNING("pybind11 is missing: Python interfaces are disabled.")
+      message (STATUS "Searching for pybind11 - not found.")
+    endif()
+  endif()
+endif()
+
 gz_configure_build(QUIT_IF_BUILD_ERRORS
   COMPONENTS log parameters)
-
```
ahcorde, line 122: discard change

python/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+
+configure_build_install_location(${BINDINGS_MODULE_NAME})
+
+install(FILES
+  src/__init__.py  
+  DESTINATION "${GZ_PYTHON_INSTALL_PATH}/transport${PROJECT_VERSION_MAJOR}"
+)
+
+if (BUILD_TESTING AND NOT WIN32)
+
```
ahcorde, line 69:
```suggestion
```

python/examples/publisher.py
```diff
@@ -0,0 +1,42 @@
+from gz.msgs10.stringmsg_pb2 import StringMsg
```
ahcorde, line 1: LICENSE

python/examples/publisher.py
```diff
@@ -0,0 +1,42 @@
...
+        count = 0
+        while True:
+          count += 1
+          vector3d_msg.x = count
+          #  if not pub_stringmsg.publish(stringmsg_msg) or not pub_vector3d.publish(vector3d_msg):
+          if not pub_vector3d.publish(vector3d_msg):
+              break
+              #  break
+
+          #  print("Publishing 'Hello' on topic [{}]".format(stringmsg_topic))
```
ahcorde, line 32: remove?

python/examples/requester.py
```diff
@@ -0,0 +1,16 @@
+from gz.msgs10.stringmsg_pb2 import StringMsg
```
ahcorde, line 1: LICENSE

python/examples/subscriber.py
```diff
@@ -0,0 +1,49 @@
+from gz.msgs10.stringmsg_pb2 import StringMsg
```
ahcorde, line 1: LICENSE

python/src/__init__.py
```diff
@@ -0,0 +1,35 @@
+from ._transport import Node as _Node
```
ahcorde, line 1: LICENSE

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,222 @@
+/*
+ * Copyright (C) 2015 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2023 Open Source Robotics Foundation
```

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,222 @@
...
+ * limitations under the License.
+ *
+*/
+
+#include <google/protobuf/message.h>
+#include <gz/transport/Node.hh>
+
+#include <pybind11/pybind11.h>
+#include <pybind11/functional.h>
+#include <pybind11/stl.h>
```
ahcorde, line 23: alphabetize

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,222 @@
...
+    py::class_<AdvertiseOptions>(
+      m, "AdvertiseOptions",
+      "A class for customizing the publication options for"
+      " a topic or service advertised")
+      .def(py::init<>())
+      .def_property("scope",
+          &AdvertiseOptions::Scope,
+          &AdvertiseOptions::SetScope,
+          "The scope used in this topic/service")
+      ;
```
ahcorde, line 48: `;` are in a weird location

python/test/requester_TEST.py
```diff
@@ -0,0 +1,22 @@
+from gz.msgs10.stringmsg_pb2 import StringMsg
```
ahcorde, line 1: LICENSE

ahcorde (changes requested):

python/CMakeLists.txt
Voldivh: Do you mean the `message("")` or the whole `if-else` statement?

python/test/requester_TEST.py
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Done.

python/src/__init__.py
Voldivh: Done.

python/examples/requester.py
Voldivh: Done.

python/examples/publisher.py
Voldivh: Done.

python/examples/publisher.py
Voldivh: Done.

python/CMakeLists.txt
Voldivh: Done.

CMakeLists.txt
Voldivh: Done.

python/examples/subscriber.py
```diff
@@ -0,0 +1,64 @@
...
+        print("Error subscribing to topic [{}]".format(topic_vector3d))
+        return
+
+    # wait for shutdown
+    try:
+      while True:
+        time.sleep(0.001)
+    except KeyboardInterrupt:
+      pass
+    # TODO(azeey) The subscriber if we quit without unsubscribing. This may not
```
ahcorde, line 57: This PR is in, remove TODO or fix this

python/src/__init__.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (C) 2015 Open Source Robotics Foundation
```
ahcorde, line 1:
```suggestion
# Copyright (C) 2023 Open Source Robotics Foundation
```

ahcorde (changes requested):

python/examples/subscriber.py
Voldivh: Done.

python/src/__init__.py
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+          const std::string &_topic)
+          {
+            std::vector<MessagePublisher> publishers;
+            std::vector<MessagePublisher> subscribers;
+            _node.TopicInfo(_topic, publishers, subscribers);
+            return std::make_pair(publishers, subscribers);
+          },
+          py::arg("topic"),
+          "Get the information about a topic")
+      .def("advertised_services", &Node::AdvertisedServices,
```
azeey, line 150: Since we don't support advertizing services, this can be removed.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+          "Get the list of services advertised by this node")
+      // send a service request using the blocking interface
+      .def("request_raw", [](
```
azeey, line 153: It's easier to check if there's parity between the C++ class and the python 
bindings if the member functions come in the order they're defined in the C++ 
class. Can you move `request_raw` to come before `topic_list`? Same with `subscribe_raw`.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+          Node &_node,
+          const std::string &_service,
+          const std::string &_request,
+          const std::string &_reqType,
+          const std::string &_repType,
+          const unsigned int &_timeout,
+          std::string &_response)
```
azeey, line 160: As I mentioned in a previous comment, we can drop the `_response` argument. 
I don't think we want to be mutating a reference anyway in pybind11. Instead, we'd 
create the response object on the stack here.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+          {
+            bool result{false};
+            result = _node.RequestRaw(_service, _request, _reqType,
+                            _repType, _timeout, _response, result);
+            return std::make_tuple(result, py::bytes(_response.c_str(), _response.size()));
```
azeey, line 165: I wonder if it's better to return with `py::make_tuple` instead.

azeey, line 165: I wonder if it's better to return with `py::make_tuple` instead?

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+          },
+          "Get the list of topics currently advertised in the network")
+      .def("topic_info", [](
+          Node &_node,
+          const std::string &_topic)
+          {
+            std::vector<MessagePublisher> publishers;
+            std::vector<MessagePublisher> subscribers;
+            _node.TopicInfo(_topic, publishers, subscribers);
+            return std::make_pair(publishers, subscribers);
```
azeey, line 146: I wonder if it's better to return with `py::make_tuple` instead?

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+    py::class_<MessageInfo>(
+      m, "MessageInfo",
+      "A class that provides information about the message received.")
+      .def(py::init<>());
+
+    auto node = py::class_<Node>(m, "Node",
+      "A class that allows a client to communicate with other peers."
+      " There are two main communication modes: pub/sub messages"
+      " and service calls")
+      .def(py::init<>())
```
azeey, line 82: This might be a follow up PR, but can you add the constructor that takes `NodeOptions`?

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+      .def("service_info", [](
+          Node &_node,
+          const std::string &_service)
+          {
+            std::vector<ServicePublisher> publishers;
+            _node.ServiceInfo(_service, publishers);
+            return publishers;
+          },
+          py::arg("service"),
+          "Get the information about a service");
```
azeey, line 192: There are a few more member functions, `Options`, `EnableStats`, and `TopicStats`. Can you add bindings for them?

python/test/pubSub_TEST.py
```diff
@@ -0,0 +1,48 @@
...
+import unittest
+
+class pubSubTEST(unittest.TestCase):
+
+    def vector3_cb(self, msg: Vector3d):
+        self.received_msg = msg.x
+
+    def setUp(self):
+        # Publisher set up
+        self.pub_node = Node()
```
azeey, line 28: Usually we put things in `setUp` things that are common to all the tests in 
this test case. If that's the intention, ignore this, otherwise, I recommend 
putting most of this code in `test_msg_callback`.

python/test/pubSub_TEST.py
```diff
@@ -0,0 +1,48 @@
...
+        self.assertTrue(self.pub_node)
```
azeey, line 29: Not sure what this is asserting

python/test/pubSub_TEST.py
```diff
@@ -0,0 +1,48 @@
...
+        self.assertTrue(self.pub)
+        self.assertFalse(self.pub.has_connections())
+
+        self.vector3d_msg = Vector3d()
+        self.vector3d_msg.x = 10
+
+        # Subscriber set up
+        self.sub_node = Node()
+        self.assertTrue(self.sub_node)
+        self.assertTrue(self.sub_node.subscribe(Vector3d, vector3d_topic, self.vector3_cb))
```
azeey, line 41: Not sure what this is asserting

python/test/pubSub_TEST.py
```diff
@@ -0,0 +1,48 @@
...
+        self.received_msg = 0
+        self.assertTrue(self.pub.has_connections())
+    
+    def test_msg_callback(self):
+        self.assertEqual(self.received_msg, 0)
+        self.assertTrue(self.pub.publish(self.vector3d_msg))
+        self.assertEqual(self.received_msg, self.vector3d_msg.x)
```
azeey, line 48: I wonder if we need to wait for the publisher to actually publish.

This also 
makes me think about data races. But reading the [pybind11 documentation](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil), it 
looks like the GIL will be locked before calling the callback.

python/test/requester_TEST.py
```diff
@@ -0,0 +1,50 @@
...
+import os
+import subprocess
+import unittest
+
+class requesterTEST(unittest.TestCase):
+
+    def setUp(self):
+        # Environment Setup
+        gz_partition = 'python_requester_test'
+        os.environ['GZ_PARTITION'] = gz_partition
```
azeey, line 28: Do we need to reset the partition after the test is done to avoid affecting 
other tests?

python/test/requester_TEST.py
```diff
@@ -0,0 +1,50 @@
...
+
+import os
+import subprocess
+import unittest
+
+class requesterTEST(unittest.TestCase):
+
+    def setUp(self):
+        # Environment Setup
+        gz_partition = 'python_requester_test'
```
azeey, line 27: I don't think using `setUp` and `tearDown` makes sense if we have just one 
test.

python/test/requester_TEST.py
```diff
@@ -0,0 +1,50 @@
...
+        os.environ['GZ_PARTITION'] = gz_partition
+        
+        # Subprocess Setup
+        cmd = f"{os.getenv('CMAKE_BINARY_DIR')}/INTEGRATION_twoProcsSrvCallReplier_aux {gz_partition}"
+        self.service_process = subprocess.Popen(cmd, shell=True)
+        
+        # Requester Setup
+        self.node = Node()
+        self.assertTrue(self.node)
```
azeey, line 36: Remove?

python/test/pubSub_TEST.py
```diff
@@ -0,0 +1,48 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from gz.msgs10.vector3d_pb2 import Vector3d
+from gz.transport13 import Node, AdvertiseMessageOptions, SubscribeOptions
+
+import unittest
+
+class pubSubTEST(unittest.TestCase):
```
azeey, line 21: Class names should start with an uppercase letter according to [PEP8](https://peps.python.org/pep-0008/#class-names).

python/test/requester_TEST.py
```diff
@@ -0,0 +1,50 @@
...
+#
+
+from gz.msgs10.int32_pb2 import Int32
+from gz.transport13 import Node
+
+import os
+import subprocess
+import unittest
+
+class requesterTEST(unittest.TestCase):
```
azeey, line 23: Class names should start with an uppercase letter according to [PEP8](https://peps.python.org/pep-0008/#class-names).

python/test/requester_TEST.py
```diff
@@ -0,0 +1,50 @@
...
+        self.response = Int32()
+        self.request.data = 100
+        self.timeout = 5000
+
+    def tearDown(self):
+        self.service_process.kill()
+
+    def test_msg_callback(self):
+        self.assertNotEqual(self.response.data, self.request.data)
+        result, response = self.node.request(self.service_name, self.request, Int32, Int32, self.timeout, self.response)
```
azeey, line 48: This line is too long. Can you reformat?

azeey (changes requested):
Looks great overall. Most of the comments should be easy to address. Some 
comments might be addressed in a follow-up PR.

python/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+)
+
+target_link_libraries(${BINDINGS_MODULE_NAME} PRIVATE
+  ${PROJECT_LIBRARY_TARGET_NAME}
+)
+
+target_compile_definitions(${BINDINGS_MODULE_NAME} PRIVATE
+  BINDINGS_MODULE_NAME=${BINDINGS_MODULE_NAME})
+
+# Deal with naming collision on Windows. It is caused by the both the import
```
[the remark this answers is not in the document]
Voldivh: I'm not completely sure what to do on this one. Should I modify the current `ARCHIVE_OUTPUT_NAME` to something like the following and remove the comment?
```CMake
set_target_properties(${BINDINGS_MODULE_NAME} PROPERTIES
  ARCHIVE_OUTPUT_NAME "_transport${PROJECT_VERSION_MAJOR}")
```

python/src/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+            deserialized_msg = msg_type()
+            deserialized_msg.ParseFromString(proto_msg)
+            callback(deserialized_msg)
+
+        return self.subscribe_raw(topic, cb_deserialize,
+                                  msg_type.DESCRIPTOR.full_name, options)
+    
+    def request(self, service, request, request_type, response_type, timeout, response):
+        result, serialized_response = self.request_raw(service, request.SerializeToString(), request_type.DESCRIPTOR.full_name,
+                                response_type.DESCRIPTOR.full_name, timeout, response.SerializeToString())
```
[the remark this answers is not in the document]
Voldivh: Yeah, I actually thought the same. We can certainly remove that argument.

CMakeLists.txt
```diff
@@ -115,6 +130,31 @@ gz_find_package(SQLite3
 #============================================================================
 # Configure the build
 #============================================================================
+
+########################################
+# Python interfaces
+if (NOT PYTHON3_FOUND)
```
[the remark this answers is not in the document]
Voldivh: Done.

python/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+# Set the build location and install location for a CPython extension
+function(configure_build_install_location _library_name)
+  # Install library for actual use
+  install(TARGETS ${_library_name}
+    DESTINATION "${GZ_PYTHON_INSTALL_PATH}/transport${PROJECT_VERSION_MAJOR}"
+  )
+endfunction()
+
+message(STATUS "Building pybind11 interfaces")
+set(BINDINGS_MODULE_NAME "_transport")
```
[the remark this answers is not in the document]
Voldivh: Sure and done.

python/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+target_link_libraries(${BINDINGS_MODULE_NAME} PRIVATE
+  ${PROJECT_LIBRARY_TARGET_NAME}
+)
+
+target_compile_definitions(${BINDINGS_MODULE_NAME} PRIVATE
+  BINDINGS_MODULE_NAME=${BINDINGS_MODULE_NAME})
+
+# Deal with naming collision on Windows. It is caused by the both the import
+# library of sdformat library and the import library of the pybind11 bindings
+# library are called sdformat13.lib.
```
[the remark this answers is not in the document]
Voldivh: Nice catch. Done.

python/src/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        def cb_deserialize(proto_msg, msg_size, msg_info):
+            deserialized_msg = msg_type()
+            deserialized_msg.ParseFromString(proto_msg)
+            callback(deserialized_msg)
+
+        return self.subscribe_raw(topic, cb_deserialize,
+                                  msg_type.DESCRIPTOR.full_name, options)
+    
+    def request(self, service, request, request_type, response_type, timeout, response):
+        result, serialized_response = self.request_raw(service, request.SerializeToString(), request_type.DESCRIPTOR.full_name,
```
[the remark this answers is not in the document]
Voldivh: Done.

python/src/__init__.py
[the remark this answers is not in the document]
Voldivh: Done.

python/src/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from ._transport import Node as _Node
+from ._transport import AdvertiseMessageOptions, SubscribeOptions
+
+class Publisher(_Node.Publisher):
+
+    def publish(self, proto_msg):
```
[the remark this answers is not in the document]
Voldivh: Sure, QQ: For the protobuf message arguments what did we agreed to use? Create a TypeVar like 
```python
from typing import TypeVar
ProtoMsgType = TypeVar('ProtoMsgType')
```

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+              const AdvertiseMessageOptions &
+          )>(&Node::Advertise),
+          py::arg("topic"),
+          py::arg("msg_type_name"),
+          py::arg("options"),
+          "Advertise a new topic. If a topic is currently advertised,"
+          " you cannot advertise it a second time (regardless of its type)")
+      .def("advertised_topics", &Node::AdvertisedTopics,
+          "Get the list of topics advertised by this node")
+      .def("subscribe", [](
```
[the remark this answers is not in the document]
Voldivh: Done.

python/test/pubSub_TEST.py
Voldivh: Yeah, my idea was to create things needed in all the test cases there. More test will come to this files so I'll take this comment into account when I'm adding them.

python/test/requester_TEST.py
Voldivh: From what I saw, the other test set this environment variable at their setup. I could be wrong about this since I didn't look at all the test and the `requester_TEST` is the last one to run. I'll take a closer look.

python/test/requester_TEST.py
Voldivh: I used the methods taking into account that more test will be incoming to the python test files. I'll take this comment into account when doing that.

python/CMakeLists.txt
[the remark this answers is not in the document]
Voldivh: Done.

python/src/__init__.py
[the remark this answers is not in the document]
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+PYBIND11_MODULE(BINDINGS_MODULE_NAME, m) {
+    py::class_<AdvertiseOptions>(
+      m, "AdvertiseOptions",
+      "A class for customizing the publication options for"
+      " a topic or service advertised")
+      .def(py::init<>())
+      .def_property("scope",
+          &AdvertiseOptions::Scope,
+          &AdvertiseOptions::SetScope,
+          "The scope used in this topic/service");
```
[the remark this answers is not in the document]
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+      .def(py::init<>())
+      .def_property_readonly("throttled",
+          &SubscribeOptions::Throttled,
+          "Whether the subscription has been throttled")
+      .def_property("msgs_per_sec",
+          &SubscribeOptions::MsgsPerSec,
+          &SubscribeOptions::SetMsgsPerSec,
+          "Set the maximum number of messages per second received per topic");
+
+    py::class_<MessageInfo>(
```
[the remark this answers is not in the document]
Voldivh: Sure and done.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Done for `request_raw`, however, I'm not sure what to do with `subscribe_raw` it already comes before `topic_list` and `request_raw`

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Done.

python/test/requester_TEST.py
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,216 @@
...
+            return _node.Subscribe(_topic, _callback, _opts);
+          },
+          py::arg("topic"),
+          py::arg("callback"),
+          py::arg("options"),
+          "Subscribe to a topic registering a callback")
+      .def("subscribe_raw", [](
+          Node &_node,
+          const std::string &_topic,
+          std::function<void(py::bytes _msgData, const size_t _size,
```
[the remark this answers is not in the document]
Voldivh: I think I don't completely understand this one. In order to create the method callback on the lambda method, we need the size of the message is there a way to get that information from the object that `py::bytes` return? I tried looking it up but I couldn't find much.

python/test/pubSub_TEST.py
Voldivh: I noticed on the current C++ test some assertions on the node creation. I was just replicating the same behavior on python. If for some reason we are unable to create the node with the bindings the assertion would fail. Let me know if you think we should remove this.

python/test/pubSub_TEST.py
Voldivh: Same comment here as in https://github.com/gazebosim/gz-transport/pull/411#discussion_r1242714434. Same logic for verifying that the subscription was created.

python/test/requester_TEST.py
Voldivh: Same comment as in https://github.com/gazebosim/gz-transport/pull/411#discussion_r1242714434

azeey:
This requires https://github.com/gazebosim/gz-msgs/pull/355.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Done. I also created the bindings for the `NodeOptions` class

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Done.

python/src/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+class Node(_Node):
+    def advertise(self, topic: str, msg_type: ProtoMsg,
+                  options=AdvertiseMessageOptions()):
+        return Publisher(
+            _Node.advertise(self, topic, msg_type.DESCRIPTOR.full_name,
+                            options))
+
+    def subscribe(self, msg_type: ProtoMsg, topic: str,
+                  callback: Callable, options=SubscribeOptions()):
+        def cb_deserialize(proto_msg, msg_size, msg_info):
```
azeey, line 40: `msg_size` is not used here, so it can be removed. This what I meant in my other comment.

python/src/transport/_gz_transport_pybind11.cc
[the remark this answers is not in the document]
azeey: I left a comment in the python callback. The size is not used there so we can 
remove it from the callback signature here and in the python code.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,291 @@
...
+          &NodeOptions::SetNameSpace,
+          "Set the node's namespace.")
+      .def_property("partition",
+          &NodeOptions::Partition,
+          &NodeOptions::SetPartition,
+          "Set the node's partition name.")
+      .def("add_topic_remap",
+          &NodeOptions::AddTopicRemap,
+          py::arg("fromTopic"),
+          py::arg("toTopic"),
```
azeey, line 97: nit: I think argument names also use the snake_case style.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,291 @@
...
+          "Add a new topic remapping.If a topic is remapped, the '_fromTopic'"
+          "topic will be renamed to '_toTopic' in any of the previous functions."
+          "Is not possible to add two remaps over the same '_fromTopic'.")
+      .def("topic_remap",[](
+          NodeOptions &_nodeOptions,
```
azeey, line 102:
```suggestion
          NodeOptions &self,
```

python/src/transport/_gz_transport_pybind11.cc
azeey: I meant the same principle of synced order between C++ and python applied to 
`subscribe_raw`. Looks like it is good now.

python/src/transport/_gz_transport_pybind11.cc
```diff
@@ -0,0 +1,291 @@
...
+          py::arg("topic"),
+          py::arg("callback"),
+          py::arg("msg_type"),
+          py::arg("options"))
+      .def_property_readonly("options", &Node::Options,
+          "Get the reference to the current node options.")
+      .def("enable_stats", &Node::EnableStats,
+          py::arg("topic"),
+          py::arg("enable"),
+          py::arg("publicationTopic"),
```
azeey, line 260: The C++ version has default arguments. Can you do the same here?

python/test/pubSub_TEST.py
azeey: Can you point me to the C++ test you're referring to. I believe the only way 
the constructor would fail is if it throws an exception and if it does throw an 
exception, the code wouldn't get to this line.

python/test/requester_TEST.py
azeey: Okay, since other tests will be running in different processes, it's probably 
okay.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: I assumed later what you meant by taking a look to the `Node.hpp` file.

python/src/transport/_gz_transport_pybind11.cc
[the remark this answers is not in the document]
Voldivh: I think I didn't explain myself appropriately. I meant that we need the size of the data in order to generate the `gz::transport::RawCallback` used by the `SubscribeRaw`. At this moment I'm implicitly using the data size retrieved by the callback on python in order to be able to create that. If I remove it from the signature I would need to get the size of the message from another place in order to create that `RawCallback`. I'm not sure if it makes sense to create an object on the lambda method (C++ bindings side) and get the size of that object as the other argument. Something like:
```C++
.def("subscribe_raw", [](
          Node &_node,
          const std::string &_topic,
          std::function<void(py::bytes _msgData,
                           const MessageInfo &_info)> &_callback,
          const std::string &_msgType,
          const SubscribeOptions &_opts)
          {
            auto _cb = [_callback](const char *_msgData,
                           const MessageInfo &_info){
                auto data_object = py::bytes(_msgData);
                return _callback(data_object, sizeof(data_object_size), _info);
            };
            return _node.SubscribeRaw(_topic, _cb, _msgType, _opts);
          },
```

python/test/pubSub_TEST.py
Voldivh: You are right. I took a quick look [here](https://github.com/Voldivh/gz-transport/blob/57561475bc2301b33ff4d503bb820693171e1caa/test/integration/twoProcsPubSub.cc#L107) and thought it was asserting the node creation but it is assertion that the method worked. I'll remove those assertions here and there.

python/src/__init__.py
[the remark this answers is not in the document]
Voldivh: Sure. Done.

python/src/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+from ._transport import Node as _Node
+from ._transport import AdvertiseMessageOptions, SubscribeOptions
+from typing import TypeVar, Callable
+
+ProtoMsg = TypeVar("ProtoMsg")
+ProtoMsgType = TypeVar("ProtoMsgType")
+
+
+class Publisher(_Node.Publisher):
+    def publish(self, proto_msg: ProtoMsg):
```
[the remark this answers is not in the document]
Voldivh: Sure.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Yep, you are right. Done.

python/src/transport/_gz_transport_pybind11.cc
Voldivh: Yep, done.

python/src/transport/_gz_transport_pybind11.cc
srmainwaring: Could you pls retain the `Copyright (C) 2022 Rhys Mainwaring` line from https://github.com/srmainwaring/gz-python/blob/main/src/pybind11/gz/transport/_gz_transport_pybind11.cc as this is based on that version.

Voldivh: Sure I don't see any problem. Wdyt @ahcorde?

python/src/__init__.py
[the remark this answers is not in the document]
Voldivh: Done.

python/src/transport/_gz_transport_pybind11.cc
srmainwaring: Thanks for adding @Voldivh - I actually meant for you to add the original note below the line for OSRF (i.e. two lines - one below for the original work and another above for the derived work - I think that's the usual practice.)

Btw - nice job, this is looking v. good. I'll try to test and give feedback against some of the examples I have.

Voldivh: @srmainwaring Done (I think :D). If you manage to have some time to test it out it would be really helpful.

Voldivh:
Added the python version example of https://github.com/gazebosim/gz-sim/tree/main/examples/standalone/multi_lrauv_race in https://github.com/gazebosim/gz-transport/pull/411/commits/2f23947262eb71c4b43c6e16705fc5565ee5f2ad using the bindings created on this PR. In order to run it follow the next steps, this takes into account that you have Gazebo installed and `gz-transport` package compiled:
1. Launch the world (This would require to have the source code of gz-sim on the `main` branch):
```bash
gz sim -r <path_to_ws>/src/gz-sim/examples/worlds/multi_lrauv_race.sdf
```
2. Update the `PYTHONPATH` variable:
```bash
export PYTHONPATH=$PYTHONPATH:<path_to_ws>/install/lib/python
```
3. Run the python script for the example:
```bash
python3 <path_to_ws>/src/gz-transport/python/examples/multi_lrauv_race.py
```

python/src/__init__.py
Voldivh: Done.

python/examples/multi_lrauv_race.py
```diff
@@ -0,0 +1,92 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+#
+# In each iteration, for each vehicle, generate a random fin angle and thrust
+# within reasonable limits, and send the command to the vehicle.
+#
+# Usage:
+#   $ python3 multi_lrauv_race
```
azeey, line 22:
```suggestion
#   $ python3 multi_lrauv_race.py
```

python/src/__init__.py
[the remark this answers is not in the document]
azeey: Nice explanation! Thanks.

python/src/transport/_gz_transport_pybind11.cc
azeey: Can you also update the docstring?

python/test/options_TEST.py
```diff
@@ -0,0 +1,69 @@
+from gz.msgs10.stringmsg_pb2 import StringMsg
+from gz.transport13 import Node, AdvertiseMessageOptions, SubscribeOptions, NodeOptions
+
+import unittest
+
+
+class PubSubTEST(unittest.TestCase):
```
azeey, line 7: Should this be renamed?

python/test/pubSub_TEST.py
```diff
@@ -0,0 +1,213 @@
...
+        self.assertEqual(topic_info[0][0].msg_type_name, 'gz.msgs.Vector3d')
+
+    # Checks that the methods to enable a topic statistics and get those stats
+    # are working as expected.
+    def test_topic_stats(self):
+        self.assertEqual(len(self.pub_node.topic_info('/statistics')[0]), 0)
+        self.assertTrue(self.pub_node.enable_stats(self.vector3d_topic, True, '/statistics', 1))
+        self.assertEqual(len(self.pub_node.topic_info('/statistics')[0]), 1)
+        topic_stats = self.pub_node.topic_stats(self.vector3d_topic)
+        self.assertEqual(topic_stats, None)
```
azeey, line 212: Should `topic_stats` be `None`? I would have expected it to have some value.

azeey (approved):
Just a few more minor comments. Otherwise, LGTM!

Would you be able to merge from main and update this to use https://github.com/gazebosim/gz-msgs/pull/355?

python/test/pubSub_TEST.py
Voldivh: I thought the same, however, take a look at the [C++ test version](https://github.com/gazebosim/gz-transport/blob/9147781be37d0ccafc1dcff741e94c9deec1d937/test/integration/statistics.cc#L71).

python/test/options_TEST.py
azeey: `OptionsTEST`?

python/test/pubSub_TEST.py
```diff
@@ -0,0 +1,213 @@
...
+        pub_2 = pub_node.advertise(string_msg_topic, StringMsg)
+        self.assertTrue(pub_2)
+        self.assertTrue(pub_2.valid())
+        self.assertFalse(pub_2.has_connections())
+
+        # Node set up
+        node = Node()
+        topics = node.topic_list()
+        self.assertTrue(topics)
+        self.assertEqual(len(topics), 2)
```
azeey, line 191: This is failing for me locally with 
```
103:         self.assertTrue(topics)
103: >       self.assertEqual(len(topics), 2)
103: E       AssertionError: 12 != 2
```

python/test/pubSub_TEST.py
azeey: Ah, okay. Mind adding a comment similar to the C++ version?

azeey (commented):
I have one test failure. Not sure if it's failing for you. Also, I know I asked 
you to rebase this to work with new message generation framework in `gz-msgs`. 
But since that's been reverted, could you also revert the change here? I plan 
to open a gz-msgs PR that uses the old message generation framework and adds 
python bindings.

python/test/options_TEST.py
Voldivh: Oh, I was mistaken on the file when I read the comment, my bad. Yeah that makes a lot of sense.

python/test/pubSub_TEST.py
Voldivh: Hmmm I'm not getting that assertion error. When I modified the msg generation framework, some test started failing (not only on python). I'll take a look after reverting that change.

python/test/pubSub_TEST.py
Voldivh: Sure, will do.

Voldivh:
> I have one test failure. Not sure if it's failing for you. Also, I know I asked you to rebase this to work with new message generation framework in `gz-msgs`. But since that's been reverted, could you also revert the change here? I plan to open a gz-msgs PR that uses the old message generation framework and adds python bindings.

No problem, just one question then, should I change this PR to target the `main` as before then?

azeey:
> > I have one test failure. Not sure if it's failing for you. Also, I know I asked you to rebase this to work with new message generation framework in `gz-msgs`. But since that's been reverted, could you also revert the change here? I plan to open a gz-msgs PR that uses the old message generation framework and adds python bindings.
> 
> No problem, just one question then, should I change this PR to target the `main` branch as before then?

Yes, target the `main` branch.

azeey:

azeey:
Needs https://github.com/gazebo-release/gz-msgs10-release/pull/6 for CI.

azeey:
I've built gz-msgs10 debians just for Jammy for testing and added a new dependency in https://github.com/gazebosim/gz-transport/pull/411/commits/e1bca5ec9b41c9569618fc83317c9a6f188e9102. It looks like tests are passing on Jenkins 🎉

python/CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+    if (pytest_FOUND)
+      add_test(NAME ${test}.py COMMAND
+        "${Python3_EXECUTABLE}" -m pytest "${CMAKE_SOURCE_DIR}/python/test/${test}.py" --junitxml "${CMAKE_BINARY_DIR}/test_results/${test}.xml")
+    else()
+      add_test(NAME ${test}.py COMMAND
+        "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/python/test/${test}.py")
+    endif()
+    set(_env_vars)
+    list(APPEND _env_vars "CMAKE_BINARY_DIR=${CMAKE_BINARY_DIR}/bin")
+    list(APPEND _env_vars "PYTHONPATH=${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/python/:${CMAKE_BINARY_DIR}/lib:$ENV{PYTHONPATH}")
```
azeey, line 91: I think the Github Actions job is failing because the tests are running without install, i.e, just `make` and `make test`. To make that work, we need change the resulting directory structure. Right now, we have
```
.
|-- bin
|-- cmake
|-- lib
|   |-- _transport.cpython-310-x86_64-linux-gnu.so
|-- test_results
`-- tutorials.md
```

But I think we want
```
.
|-- bin
|-- cmake
|-- lib
|   |-- gz
|   |   |-- transport13
|   |   |   |-- __init__.py
|   |   |   |-- _transport.cpython-310-x86_64-linux-gnu.so
|-- test_results
`-- tutorials.md
 ```

azeey: I'm not sure how we can do that exactly. The alternative, and the quicker way is to add a `.github/ci/after_make.sh` similar to https://github.com/azeey/gz-msgs/blob/9b14d178c6673ef9b56c959b52868e54cc1199fe/.github/ci/after_make.sh.

Voldivh: I added it on https://github.com/gazebosim/gz-transport/pull/411/commits/3bfb36500c71cb43ae2d46368d1903e0641f767d, however, CI is still failing. I noticed an error that it was not able to find `python3-gz-msgs10` package.

azeey:
https://github.com/gazebo-release/gz-msgs10-release/pull/6 is merged and I've built the debs for Focal and Jammy, so I'll run tests again.

azeey:

python/CMakeLists.txt
azeey: I've resolved the packaging issue, but there was a type in the name of the `after_make.sh` file (fixed inhttps://github.com/gazebosim/gz-transport/pull/411/commits/5371b41adbb8af9d36ab30c77db1e796a0551b82). Now all CI is passing!!

python/test/pubSub_TEST.py
azeey: Okay, this is no longer failing for me. Probably something wrong with my local setup last time I tried.

azeey (approved):
@ahcorde All your feedback has been resolved. Do you want to take another look?

ahcorde (approved):

PR #425, opened 2023-07-27
Allow transport to generate messages

mjcarroll:
Make use of https://github.com/gazebosim/gz-msgs/pull/368

### DIFF
--- a/proto/CMakeLists.txt
+++ b/proto/CMakeLists.txt
@@ -1,0 +1,23 @@
+set(proto_files
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/discovery.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_declaration.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_declarations.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_error.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_name.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_value.proto
+)
+
+gz_msgs_generate_messages(
+  TARGET gz-transport${PROJECT_VERSION_MAJOR}-msgs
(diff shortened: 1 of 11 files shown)


proto/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_declaration.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_declarations.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_error.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_name.proto
+  ${PROJECT_SOURCE_DIR}/proto/gz/msgs/parameter_value.proto
+)
+
+gz_msgs_generate_messages(
+  TARGET gz-transport${PROJECT_VERSION_MAJOR}-msgs
```
azeey, line 12: We discussed making the generated library a component so it can be used by third-party applications without having to link against the containing library. Is that still possible?

proto/gz/msgs/discovery.proto
```diff
@@ -0,0 +1,181 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+syntax = "proto3";
+package gz.msgs;
```
azeey, line 19: Okay, so we're keeping the package name `gz.msgs`. I think it's okay, but would be a little confusing for future users.

proto/CMakeLists.txt
mjcarroll: You should be able to use the msgs component without using the core library for each package that has it.

proto/gz/msgs/discovery.proto
mjcarroll: It's a little confusing, but also limits the number of changes that need to be made.  I think if we want, we can then alter the package names in a later revision, but it makes the diff quite large here.

PR #457, opened 2023-10-31
Remove duplicated functionality from test_config

mjcarroll:
* Use `gz::utils::env` family of functions rather than the OS equivalents (which are also missing on Windows)
* Use `gz::utils` disable macros rather than custom ones.
* Remove `std::filesystem` logic, as we are only supporting c++17 and above here.

### DIFF
--- a/src/Discovery_TEST.cc
+++ b/src/Discovery_TEST.cc
@@ -27,20 +27,10 @@
 #include "gz/transport/Publisher.hh"
 #include "gz/transport/TransportTypes.hh"
 #include "gz/transport/Uuid.hh"
-#include "test_config.hh"
 
-// Temporarily introduce a "DISABLED_ON_LINUX" macro.
-// It currently does not exist upstream.
-// This can be removed when it is in upstream gz-utils
-// or the discovery WrongGzIp test passes on linux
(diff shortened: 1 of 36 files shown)


src/Discovery_TEST.cc
```diff
@@ -540,31 +531,24 @@ TEST(DiscoveryTest, TestActivity)
...
-#if defined __linux__
-  #define TEST_NAME DISABLED_WrongIgnIp
-#else
-  #define TEST_NAME WrongIgnIp
-#endif  // defined __linux__
-
 //////////////////////////////////////////////////
 /// \brief Check that a wrong GZ_IP value makes HostAddr() to return 127.0.0.1
-TEST(DiscoveryTest, TEST_NAME)
+TEST(DiscoveryTest, GZ_UTILS_TEST_DISABLED_ON_LINUX(WrongIgnIp))
```
ahcorde, line 536:
```suggestion
TEST(DiscoveryTest, GZ_UTILS_TEST_DISABLED_ON_LINUX(WrongGzIp))
```

ahcorde (approved):
Just a minor change in test name

PR #470, opened 2024-01-18
Fix issue #468

### LINKED ISSUE #468: Service requester starting before responder causes timeout when both are in the same process
## Environment
* OS Version: Ubuntu 22.04
* Source or binary build? source, 764a04b

## Description
* Expected behavior: When the node of a service requester is started before the node of a responder, the service request should succeed as long as the responder is started before the requester's timeout.
* Actual behavior: Currently, if the requester and responder are in different processes, the expected behavior occurs, but if they are in the same process, the requester will always time out.

  
## Steps to reproduce

1. Checkout the `flaky_service` branch from https://github.com/azeey/gz-transport/
2. Build `examples/requester.cc` and run it.

The example shows that if the responder is started with a 2s delay in a separate thread, the request fails.

caguero:
# 🦟 Bug fix

Fixes #468 

## Summary

As described in issue #468, the case of a `Request()` before an `Advertise()` when both calls are within the same process wasn't considered. This patch makes it work as in the case where requester and responser are in different processes (which was already working).

## Checklist
- [x] Added tests

### DIFF
--- a/include/gz/transport/Discovery.hh
+++ b/include/gz/transport/Discovery.hh
@@ -306,6 +306,8 @@ namespace gz
       /// (e.g. if the discovery has not been started).
       public: bool Advertise(const Pub &_publisher)
       {
+        DiscoveryCallback<Pub> cb;
+
         {
           std::lock_guard<std::mutex> lock(this->mutex);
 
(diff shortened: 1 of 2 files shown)


iche033 (approved):

include/gz/transport/Discovery.hh
```diff
@@ -317,6 +317,9 @@ namespace gz
             return false;
         }
 
+        if (this->connectionCb)
+          this->connectionCb(_publisher);
```
azeey, line 321: This fixes the test case for me, but would it handle a possible race condition where the requester makes the request right after this line executes?

caguero: I've included the callback within the `lock_guard` for consistency with the `Discover()` function. [ec324d2](https://github.com/gazebosim/gz-transport/pull/470/commits/ec324d2577ecc37cf07b29dd01876f8d36eeb29a)

azeey (approved):

Issue #475, opened 2024-01-26
Optional `--timeout` flag in `gz service -r` 


azeey:
## Desired behavior
Currently, the `--timeout` flag is required in `gz service -r`, but it would improve usability if we had a default timeout and make the flag optional

## Alternatives considered
n/a

## Implementation suggestion
Use a default timeout of `1.0 s`.

## Additional context
n/a

Issue #474, opened 2024-01-26
`gz service -r` should not require specifying a dummy message if the request type is `Empty`


azeey:
## Desired behavior
Currently, `gz service -r` requires specifying `unused: true` on requests with an `Empty` request type. This is not well documented and seems unintuitive.  It would be great if the user is able to provide an empty message (`gz service -r ' ' ...`) or no value at all (`gz service -r ...`). 

Together with #473, it would be nice to be able to run
```
gz service -s /gazebo/worlds -r --timeout 2000
```
instead of
```
gz service -s /gazebo/worlds --reqtype gz.msgs.Empty --reptype gz.msgs.StringMsg_V -r 'unused: true' --timeout 2000
```

## Alternatives considered
n/a

## Implementation suggestion
Handle the case where `reqtype` is `Empty` so that an empty message is considered valid.

## Additional context
n/a

caguero:
@azeey , What do you think about allowing `-r ""`? The option `-r` is configured as a parameter that accepts a value. Not sure if we can make the value optional.

mjcarroll:
I believe you can add `->expected(0,1)` to make it take 0 or 1 arguments

azeey:
I'm okay with `-r ""`, but if @mjcarroll's suggestion works, that would be even better.

caguero:
> I believe you can add `->expected(0,1)` to make it take 0 or 1 arguments

Thanks, it works!

Issue #473, opened 2024-01-26
`gz service -r` should not require `--reqtype` and `--reptype`


azeey:
## Desired behavior
`gz service -r` automatically determines the Request and Response message types.  So instead of:
```
gz service -s /gazebo/worlds --reqtype gz.msgs.Empty --reptype gz.msgs.StringMsg_V -r 'unused: true' --timeout 2000
```
I'd like to run:
```
gz service -s /gazebo/worlds  -r 'unused: true' --timeout 2000
```

## Alternatives considered
n/a

## Implementation suggestion
`gz service -i` is able to retrieve the request and response types so it should be possible to use the same mechanism when making service calls via `gz service -r`.

## Additional context
n/a

sam-ovalle:
Ill be looking over this one with @haresh-seenivasagan (github.com/haresh-seenivasagan)

ielson:
Hey @azeey PR is ready for review:
https://github.com/gazebosim/gz-transport/pull/829

PR #477, opened 2024-02-02
Oneway service request from the command line

### LINKED ISSUE #97: Call one-way requester / responder from command line
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-transport/issues/97)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
It currently isn't possible to call a service which has no request or one that has no response from the command line, because it requires that all fields are set.

caguero:
# 🦟 Bug fix

Partially fixes #97

## Summary

This patch makes it possible to request a one-way service request from the command line. Before, it was possible passing `--reptype gz.msgs.Empty` but always getting the message `Service call timed out` as an output (because the service callback wasn't answering anything, which was expected).

## How to test it?

You can compile the examples, and then:

1. Launch the one-way responser:
```
./responser_oneway
```

2. Request the service using the two-way request and verify that you don't get the timeout:
```
gz service -s /oneway --reqtype gz.msgs.StringMsg --reptype gz.msgs.Empty --req 'data: "Hello"' --timeout 1000
```

3. Use the new one-way request and verify that you don't get the timeout message:
```
gz service -s /oneway --reqtype gz.msgs.StringMsg --req 'data: "Hello"'
``` 

Note that both requests should reach the responser:
```
Request received: [Hello]
Request received: [Hello]
```

## Checklist
- [x] Added tests


azeey:
Would it be possible to detect that the response type is `Empty` instead of relying on the timeout? I like the idea of making some of the arguments optional (see https://github.com/gazebosim/gz-transport/issues/475, https://github.com/gazebosim/gz-transport/issues/474, https://github.com/gazebosim/gz-transport/issues/473), but there is a lot of existing documentation that uses the ` --reptype gz.msgs.Empty`  and `--timeout` arguments. It would be nice if those commands didn't timeout as well.

caguero:
> 26d669b

Sounds good to me. Then, now both:

```
gz service -s /oneway --reqtype gz.msgs.StringMsg --reptype gz.msgs.Empty --req 'data: "Hello"' --timeout 10
```
and
```
gz service -s /oneway --reqtype gz.msgs.StringMsg --req 'data: "Hello"'
```

shoudn't generate timeouts.

caguero:
@j-rivero , do you know why the same job (Ubuntu CI / Ubuntu Jammy CI) can pass and fail at the same time? It seems that `GZ_PATH` and `g_gzVersion` is undefined.

azeey:
There have been some changes in https://github.com/gazebosim/gz-transport/pull/481 that affect this PR. I suggest merging from `gz-transport13`.

caguero:
Good to go again!

src/cmd/gz.hh
```diff
@@ -44,24 +44,26 @@ extern "C" void cmdServiceList();
...
-                                                       const char *_msgData);
+                            const char *_msgType,
+                            const char *_msgData);
 
 /// \brief External hook to execute 'gz service -r' from the command line.
 /// \param[in] _service Service name.
 /// \param[in] _reqType Message type used in the request.
 /// \param[in] _repType Message type used in the response.
 /// \param[in] _timeout The request will timeout after '_timeout' ms.
+///                     If -1 is used, the request will be one-way and _repType
```
azeey, line 55: Users might expect a timeout of -1 as "wait forever", which would be the opposite of this. I would say just checking if the message type is `Empty` is sufficient, but if you think we should have the timeout check as well, we should definitely update the help message of the CLI to mention this.

azeey (approved):
LGTM! Just one comment about the timeout.

src/cmd/gz.hh
caguero: I think it's fine just checking the response type. I updated it in [687edcc](https://github.com/gazebosim/gz-transport/pull/477/commits/687edcc8fc79919920f8b824d952d93472f87a49).

PR #478, opened 2024-02-03
Merge 12 ➡️ 13

caguero:
# ➡️ Forward port

Port `gz-transport12` to `gz-transport13`

Branch comparison: https://github.com/gazebosim/gz-transport/compare/gz-transport13...gz-transport12

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)

PR #481, opened 2024-02-12
Re-enable tests of bash completion functions for gz

mjcarroll:
Follow-up from #478


caguero (commented):
Thanks!

`homebrew` looks unhappy

mjcarroll:
> homebrew looks unhappy

I'm not sure these tests worked on homebrew before, I will limit them to just Linux.

mjcarroll:
Trying something slightly different and starting with `/usr/bin/bash` and falling back to `/bin/bash` if that isn't available.

mjcarroll:
@scpeters did a brief inspection for me on macOS and things didn't seem to work, likely because of some security mechanism of the operating system.  Since the value of this test is mainly in checking the flags for consistency and keeping them in sync between the completion scripts and the executables, running cross-platform isn't really necessary here.

caguero (approved):

PR #484, opened 2024-03-18
Use `std::shared_ptr` for `gz::transport::NodeShared`

azeey:
# 🦟 Bug fix

## Summary
Currently, `NodeShared` is never destroyed once it's created. This makes
it hard for writing tests that do not interfere with each other. This
patch uses a reference counted smart pointer (`std::shared_ptr`) to keep
track of `NodeShared` instances, so that it gets properly destroyed when
when all `Node` instances, which themselves contain `NodeShared`
instances are destroyed.

This was brought up in https://github.com/gazebosim/gz-sim/pull/2334. The `INTEGRATION_triggered_publisher` test fails reliably in my local tests without this gz-transport PR.

~TODO: This is probably not threadsafe. It's possible to lock a mutex before checking the weak_ptr, but that would make the call to `NodeSHared::Instance` slower than the previous implementation since it was using `std::shared_mutex`.~

## Checklist
- [ ] Added tests


iche033 (approved):
the approach looks good to me. 

Maybe @caguero can take a look as well and see if you are ok with this.

ahcorde (approved):

caguero (approved):

azeey:
I can go ahead and merge this, but I think we should make a prerelease and test gz-sim before making a full release. @iche033  or @caguero can I ask one of you to do that please?

PR #497, opened 2024-04-30
Allow programmatic configuration of unicast relays.

mbeards:
This change allows users to configure relays from code without having to `setenv(GZ_RELAY)`.

# 🎉 New feature

## Summary
This change allows users to configure relays from code without having to `setenv(GZ_RELAY)`.

## Test it
This can be tested by starting a Node across a boundary where multicast isn't enabled, and setting the relay IP via NodeOptions.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/transport/Discovery.hh
+++ b/include/gz/transport/Discovery.hh
@@ -740,6 +740,26 @@ namespace gz
         }
       }
 
+      /// \brief Register a new relay address.
+      /// \param[in] _ip New IP address.
+      public: void AddRelayAddress(const std::string &_ip)
+      {
+        // Sanity check: Make sure that this IP address is not already saved.
+        for (auto const &addr : this->relayAddrs)
+        {
+          if (addr.sin_addr.s_addr == inet_addr(_ip.c_str()))
+            return;
+        }
+
+        sockaddr_in addr;
+        memset(&addr, 0, sizeof(addr));
+        addr.sin_family = AF_INET;
+        addr.sin_addr.s_addr = inet_addr(_ip.c_str());
+        addr.sin_port = htons(static_cast<u_short>(this->port));
+
+        this->relayAddrs.push_back(addr);
+      }
+
       /// \brief Broadcast periodic heartbeats.
       private: void UpdateHeartbeat()
       {
@@ -1418,26 +1438,6 @@ namespace gz
         }
 
         return true;
-      }
-
-      /// \brief Register a new relay address.
-      /// \param[in] _ip New IP address.
-      private: void AddRelayAddress(const std::string &_ip)
-      {
-        // Sanity check: Make sure that this IP address is not already saved.
-        for (auto const &addr : this->relayAddrs)
-        {
-          if (addr.sin_addr.s_addr == inet_addr(_ip.c_str()))
-            return;
-        }
-
-        sockaddr_in addr;
-        memset(&addr, 0, sizeof(addr));
-        addr.sin_family = AF_INET;
-        addr.sin_addr.s_addr = inet_addr(_ip.c_str());
-        addr.sin_port = htons(static_cast<u_short>(this->port));
-
-        this->relayAddrs.push_back(addr);
       }
 
       /// \brief Default activity interval value (ms.).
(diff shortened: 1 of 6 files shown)


include/gz/transport/Discovery.hh
```diff
@@ -740,6 +740,26 @@ namespace gz
...
+      {
+        // Sanity check: Make sure that this IP address is not already saved.
+        for (auto const &addr : this->relayAddrs)
+        {
+          if (addr.sin_addr.s_addr == inet_addr(_ip.c_str()))
+            return;
+        }
+
+        sockaddr_in addr;
+        memset(&addr, 0, sizeof(addr));
```
ahcorde, line 755: include `<cstring>`

src/NodeOptions.cc
```diff
@@ -136,3 +136,15 @@ bool NodeOptions::TopicRemap(const std::string &_fromTopic,
 
   return topicIt != this->dataPtr->topicsRemap.end();
 }
+
+//////////////////////////////////////////////////
+bool NodeOptions::SetRelays(const std::vector<std::string>& _relayIPs) {
```
ahcorde, line 141: include `<vector>`

include/gz/transport/Discovery.hh
mbeards: This file already includes <string.h> - should I swap the include?

caguero (changes requested):
My main problem with this patch is that so far `NodeOptions` are all options that affect a single node. However, adding relays will have a global impact in all the nodes.

I understand that we don't have a way to set global options right now. Perhaps an alternative could be to not use `NodeOptions` and just create an `AddGlobalRelay()` function in Node()` to preserve the semantics of `NodeOptions`? What do you think?

include/gz/transport/Discovery.hh
```diff
@@ -1420,26 +1440,6 @@ namespace gz
         return true;
       }
 
-      /// \brief Register a new relay address.
```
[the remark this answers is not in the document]
mbeards: I moved it up to sit with the other public functions. ("Group similar declarations together, placing public parts earlier." https://google.github.io/styleguide/cppguide.html#Declaration_Order)

mbeards:
> My main problem with this patch is that so far `NodeOptions` are all options that affect a single node. However, adding relays will have a global impact in all the nodes.

Agreed that this is weird / gross. I guess ideally I'd want to have accessors for the global Discovery instances, but threading that through seems difficult.

> I understand that we don't have a way to set global options besides using environment variables right now. Perhaps an alternative could be to not use `NodeOptions` and just create an `AddGlobalRelay()` / `GlobalRelays()` functions in `Node` to preserve the semantics of `NodeOptions`? What do you think?

You're thinking something like the following?
```
class Node {
  // ...
  public: static void AddGlobalRelay(const std::string&);
  public: static std::vector<std::string>& Relays();
  // ...
}
```

caguero:
> > My main problem with this patch is that so far `NodeOptions` are all options that affect a single node. However, adding relays will have a global impact in all the nodes.
> 
> Agreed that this is weird / gross. I guess ideally I'd want to have accessors for the global Discovery instances, but threading that through seems difficult.
> 
> > I understand that we don't have a way to set global options besides using environment variables right now. Perhaps an alternative could be to not use `NodeOptions` and just create an `AddGlobalRelay()` / `GlobalRelays()` functions in `Node` to preserve the semantics of `NodeOptions`? What do you think?
> 
> You're thinking something like the following?
> 
> ```
> class Node {
>   // ...
>   public: static void AddGlobalRelay(const std::string&);
>   public: static std::vector<std::string>& Relays();
>   // ...
> }
> ```

Probably `AddGlobalRelay()` and `GlobalRelays()`.

mbeards:
> > You're thinking something like the following?
> > ```
> > class Node {
> >   // ...
> >   public: static void AddGlobalRelay(const std::string&);
> >   public: static std::vector<std::string>& Relays();
> >   // ...
> > }
> > ```
> 
> Probably `AddGlobalRelay()` and `GlobalRelays()`.

Done in https://github.com/gazebosim/gz-transport/pull/498

PR #498, opened 2024-05-06
Allow programmatic configuration of unicast relays.

mbeards:
This change allows users to configure relays from code without having to `setenv(GZ_RELAY)`.

# 🎉 New feature

## Summary
This change allows users to configure relays from code without having to `setenv(GZ_RELAY)`.

Supersedes https://github.com/gazebosim/gz-transport/pull/497

## Test it
This can be tested by starting a Node across a boundary where multicast isn't enabled, and setting the relay IP via NodeOptions.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/transport/Discovery.hh
+++ b/include/gz/transport/Discovery.hh
@@ -64,7 +64,6 @@
 #include <map>
 #include <memory>
 #include <mutex>
-#include <shared_mutex>
 #include <string>
 #include <thread>
 #include <vector>
@@ -745,7 +744,7 @@ namespace gz
       /// \param[in] _ip New IP address.
       public: void AddRelayAddress(const std::string &_ip)
       {
-        std::lock_guard<std::shared_mutex> lock(this->relayAddrsMutex);
+        std::lock_guard<std::mutex> lock(this->mutex);
         // Sanity check: Make sure that this IP address is not already saved.
         for (auto const &addr : this->relayAddrs)
         {
@@ -768,7 +767,7 @@ namespace gz
       {
         std::vector<std::string> result;
 
-        std::shared_lock<std::shared_mutex> lock(this->relayAddrsMutex);
+        std::lock_guard<std::mutex> lock(this->mutex);
 
         for (auto const &addr : this->relayAddrs) {
           result.push_back(inet_ntoa(addr.sin_addr));
@@ -1291,7 +1290,7 @@ namespace gz
         if (_msg.SerializeToArray(buffer + sizeof(msgSize), msgSize))
         {
           // Send the discovery message to the unicast relays.
-          std::shared_lock<std::shared_mutex> lock(this->relayAddrsMutex);
+          std::lock_guard<std::mutex> lock(this->mutex);
 
           for (const auto &sockAddr : this->relayAddrs)
           {
@@ -1553,9 +1552,6 @@ namespace gz
 
       /// \brief Collection of socket addresses used as remote relays.
       private: std::vector<sockaddr_in> relayAddrs;
-
-      /// \brief Mutex to guarantee exclusive write access to relayAddrs.
-      private: mutable std::shared_mutex relayAddrsMutex;
 
       /// \brief Mutex to guarantee exclusive access between the threads.
       private: mutable std::mutex mutex;
--- a/include/gz/transport/Discovery.hh
+++ b/include/gz/transport/Discovery.hh
@@ -64,6 +64,7 @@
 #include <map>
 #include <memory>
 #include <mutex>
+#include <shared_mutex>
 #include <string>
 #include <thread>
 #include <vector>
@@ -744,6 +745,7 @@ namespace gz
       /// \param[in] _ip New IP address.
       public: void AddRelayAddress(const std::string &_ip)
       {
+        std::lock_guard<std::shared_mutex> lock(this->relayAddrsMutex);
         // Sanity check: Make sure that this IP address is not already saved.
         for (auto const &addr : this->relayAddrs)
         {
@@ -765,6 +767,8 @@ namespace gz
       public: std::vector<std::string> RelayAddresses() const
       {
         std::vector<std::string> result;
+
+        std::shared_lock<std::shared_mutex> lock(this->relayAddrsMutex);
 
         for (auto const &addr : this->relayAddrs) {
           result.push_back(inet_ntoa(addr.sin_addr));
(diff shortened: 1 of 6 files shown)


mbeards:
Friendly ping @caguero

caguero (commented):
By making `Discovery::AddRelayAddress()` public, I think we now have risk for potential race conditions, as `this->relayAddrs` can be modified from a user thread (via `Node`) and the discovery thread can read the same variable via `SendUnicast()`.

mbeards:
> By making `Discovery::AddRelayAddress()` public, I think we now have risk for potential race conditions, as `this->relayAddrs` can be modified from a user thread (via `Node`) and the discovery thread can read the same variable via `SendUnicast()`.

Should I add a new mutex to guard relayAddrs or just use the existing `Discovery::mutex`?

caguero:
> > By making `Discovery::AddRelayAddress()` public, I think we now have risk for potential race conditions, as `this->relayAddrs` can be modified from a user thread (via `Node`) and the discovery thread can read the same variable via `SendUnicast()`.
> 
> Should I add a new mutex to guard relayAddrs or just use the existing `Discovery::mutex`?

I think you can reuse `Discovery::mutex`.

mbeards:
> > > By making `Discovery::AddRelayAddress()` public, I think we now have risk for potential race conditions, as `this->relayAddrs` can be modified from a user thread (via `Node`) and the discovery thread can read the same variable via `SendUnicast()`.
> > 
> > 
> > Should I add a new mutex to guard relayAddrs or just use the existing `Discovery::mutex`?
> 
> I think you can reuse `Discovery::mutex`.

Is there a performance concern to worry about re: serializing the user thread + discovery thread in the presence of unicast relays? Every unicast write would need to acquire that mutex.

mbeards:
Thoughts on this test failure https://build.osrfoundation.org/job/gz_transport-pr-win/99/ ? I'm having a hard time figuring out how this PR would trigger behavior change on those tests.

caguero:
> > > > By making `Discovery::AddRelayAddress()` public, I think we now have risk for potential race conditions, as `this->relayAddrs` can be modified from a user thread (via `Node`) and the discovery thread can read the same variable via `SendUnicast()`.
> > > 
> > > 
> > > Should I add a new mutex to guard relayAddrs or just use the existing `Discovery::mutex`?
> > 
> > 
> > I think you can reuse `Discovery::mutex`.
> 
> Is there a performance concern to worry about re: serializing the user thread + discovery thread in the presence of unicast relays? Every unicast write would need to acquire that mutex.



> > > > By making `Discovery::AddRelayAddress()` public, I think we now have risk for potential race conditions, as `this->relayAddrs` can be modified from a user thread (via `Node`) and the discovery thread can read the same variable via `SendUnicast()`.
> > > 
> > > 
> > > Should I add a new mutex to guard relayAddrs or just use the existing `Discovery::mutex`?
> > 
> > 
> > I think you can reuse `Discovery::mutex`.
> 
> Is there a performance concern to worry about re: serializing the user thread + discovery thread in the presence of unicast relays? Every unicast write would need to acquire that mutex.

I'd expect a bit of impact if unicast is used because the mutex will be acquired when calling `DispacthDiscoveryMsg()`. It shouldn't make a big difference.

caguero:
> Thoughts on this test failure https://build.osrfoundation.org/job/gz_transport-pr-win/99/ ? I'm having a hard time figuring out how this PR would trigger behavior change on those tests.

`ReplayTest` looks unrelated.

include/gz/transport/Discovery.hh
```diff
@@ -1535,6 +1554,9 @@ namespace gz
       /// \brief Collection of socket addresses used as remote relays.
       private: std::vector<sockaddr_in> relayAddrs;
 
+      /// \brief Mutex to guarantee exclusive write access to relayAddrs.
+      private: mutable std::shared_mutex relayAddrsMutex;
```
caguero, line 1558: Why do you prefer to create a new mutex?

mbeards: Just to get to narrower locking. I can change it back to use Discovery::mutex if you'd prefer.

caguero: I think so, having multiple mutexes always makes me a bit more nervous as it's a bit harder to reason about deadlocks.

mbeards: Done.

mbeards:
Friendly ping @caguero .

caguero (commented):
Minor comments.

mbeards:
> Minor comments.

Done.

caguero (approved):

caguero (commented):
Not sure why the ABI checker doesn't catch it but I think that changing the access right (from private to public in our case) in `Discovery::AddRelayAddress()` breaks ABI.

I think the solution is to create a new public version of the function that calls the existing one.

src/Node_TEST.cc
```diff
@@ -2341,6 +2341,31 @@ TEST(NodeTest, statistics)
   EXPECT_EQ(std::nullopt, node.TopicStats("/test"));
 }
 
+//////////////////////////////////////////////////
+/// \brief Test adding and querying relays
+TEST(NodeTest, relay) {
+  transport::Node node;
+
+  // Make sure the relay we're about to add hasn't already been configured.
+  const std::vector<std::string> relays_before_add = node.GlobalRelays();
```
caguero, line 2350: Nit: CamelCase all variables.

mbeards:
> Not sure why the ABI checker doesn't catch it but I think that changing the access right (from private to public in our case) in `Discovery::AddRelayAddress()` breaks ABI.
> 
> I think the solution is to create a new public version of the function that calls the existing one.

Let's just target main?

mjcarroll:
> Not sure why the ABI checker doesn't catch it but I think that changing the access right (from private to public in our case) in Discovery::AddRelayAddress() breaks ABI.

I think it's actually fine on Linux, it's only an issue with MSVC where the symbol mangling includes the access rights

caguero:
@osrf-jenkins run tests

caguero (approved):

caguero:
Not sure why this `gz_transport-ci-pr_any-jammy-amd64` job is stuck.

azeey:
Our CI was recently changed so that `main` branches only running tests on Noble. But the branch protection for gz-transport wasn't updated. Should be good now.

azeey:
We do need to update the GitHub workflow to use Noble . @caguero would you mind doing that when you get a chance?

caguero:
> We do need to update the GitHub workflow to use Noble . @caguero would you mind doing that when you get a chance?

Thanks @scpeters https://github.com/gazebosim/gz-transport/pull/504

iche033:
@mergifyio backport gz-transport13

PR #512, opened 2024-08-02
Add an explicit dependency on cppzmq-dev

azeey:
# 🦟 Bug fix

## Summary
In Jammy, the zmq C++ headers were included in libzmq-dev. However, in Noble, the headers are now in a separate package, `cppzmq-dev`, and that package is recommended by libzmq-dev. This shouldn't affect most users, but anywhere these dependencies are installed with `--no-install-recommends` will now break because it won't have cppzmq-dev installed. This patch just adds an explicit dependency to avoid the problem.

## Checklist
- [ ] Added tests

### DIFF
--- a/.github/ci/packages-noble.apt
+++ b/.github/ci/packages-noble.apt
@@ -1,0 +1,1 @@
+cppzmq-dev
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -1,3 +1,4 @@
+cppzmq-dev
 libgz-cmake4-dev
 libgz-math8-dev
 libgz-msgs11-dev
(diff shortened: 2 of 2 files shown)


iche033:
I think this needs to be done in a separate `packages-noble.apt` file?

azeey:
> I think this needs to be done in a separate `packages-noble.apt` file?

Done in https://github.com/gazebosim/gz-transport/pull/512/commits/0135bfb60a992706b24b24cd25a92a653a55e2cb. We should be replacing all the Jammy CI jobs with Noble, but we were having issues with running the code coverage tools in https://github.com/gazebosim/gz-fuel-tools/pull/427

iche033 (approved):

PR #533, opened 2024-09-12
Windows fixes for statistics tutorial

j-rivero:
Tested in https://github.com/gazebosim/gazebo_test_cases/issues/1566

### DIFF
--- a/tutorials/23_topic_statistics.md
+++ b/tutorials/23_topic_statistics.md
@@ -79,6 +79,9 @@ For running the example, build the binaries in the example directory:
 git clone https://github.com/gazebosim/gz-transport -b gz-transport14
 cd gz-transport/example
 cmake -S . -B build
+# For UNIX
+cmake --build build --parallel
+# For Windows
 cmake --build build --config release --parallel
 ```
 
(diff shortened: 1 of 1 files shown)


caguero (approved):

tutorials/23_topic_statistics.md
```diff
@@ -73,9 +73,23 @@ if (!node.EnableStats(topic, true, "/my_stats", 100))
...
 
-If you have the Gazebo Transport sources with the example programs built,
-then you can test topic statistics by following these steps.
+For running the example, build the binaries in the example directory:
+
+```
+git clone https://github.com/gazebosim/gz-transport -b gz-transport14
+cd gz-transport/example
+cmake -S . -B build
+cmake --build build --config release --parallel
```
azeey, line 82: Is the `config` necessary? Does it work with compilers that don't have multiple configurations like gcc?

j-rivero: It is. `CMAKE_BUILD_TYPE` passed at configuring time is not working on multi-configuration generators like MSVC. It has to be decided at building time. See https://stackoverflow.com/a/64719718.

>  Does it work with compilers that don't have multiple configurations like gcc?

Worked for me testing things yesterday. I tried not to duplicate commands for UNIX/Win and use this same command. Might occur that it does not work under some configurations/compilers. Let me change that to be sure that we are not breaking anyone.

j-rivero:
Made a merge with updates in [490975b](https://github.com/gazebosim/gz-transport/pull/533/commits/490975bf770db5962e33e5f8c87d0243b0cbecf7). @caguero / @azeey if you can give it a second look? Thanks!

azeey (approved):

PR #535, opened 2024-09-13
Multiple  NodeShared per process

### LINKED ISSUE #531: Ability to set GZ_DISCOVERY_* variables from the Python API
## Desired behaviour 

Currently, for a single process, python interpreter doesn’t perform dlclose after opening any cython library until the process terminates. This produces issues when the variables have global visibility, which is the default for pybind. We can set visibility for these variables within only the objects scope by passing appropriate flags in the pybind module.

My specific problem has to do with the Node creation. I want to instantiate multiple Nodes in isolated network in a single process. Currently due to the mentioned problem, this is impossible because the Node will stay with the env vars found at first import.  We cannot delete references to the module from sys.modules and obtain desired behaviour.

I would like to work on this as well, to contribute to the project.

## Additional context

Multiple isolated Nodes in a single Python process.

AmalDevHaridevan:
# 🎉 New feature

Closes #531

## Summary

Currently the transport architecture restricts a single `NodeShared` object per process. This is achieved by mapping process id to an instance of `NodeShared` class. However, recently, I had some use cases where I need to isolate several nodes, each with their own `DISCOVERY` variables, such as `GZ_DISCOVERY_MSG_PORT`,`GZ_DISCOVERY_SRV_PORT`, `GZ_DISCOVERY_MULTICAST_IP`,`GZ_PARTITION`, in a single process. Due to the limitation mentioned above, this is not possible because
`DISCOVERY` is tied to a process, and when instantiating multiple Nodes per process, all of them obtain the same NodeShared object and hence the same `DISCOVERY` variables. 

To overcome this, one approach is to also use a `thread_id` to create a unique identifier that maps to a `NodeShared` object. This pins down the uniqueness of `NodeShared` object to each thread, not the process. The proposed modification performs this by the following logic:

``` 
// Map for obtaining shared node
static std::unordered_map<std::string, NodeShared*> nodeSharedMap;


 auto pid = getProcessId();
  // Get current thread ID, so we can create multiple NodeShared per Process, per thread
  auto tid = std::this_thread::get_id();
  // Create a unique id using both pid and tid
  std::stringstream ss;
  ss << tid;
  std::stringstream unique_ss;
  unique_ss << std::hex << pid << "-" << ss.str();
  auto uid = unique_ss.str();

 // create or access the shared node
 nodeSharedMap.find(uid) ; //access
// OR
 nodeSharedMap.insert(uid, new NodeShared) ; //create
```
This allows specification of `DISCOVERY` variables per thread.
 
## Test it

For testing the proposed improvement, I added two test files `nodeDiscoveryPubs_TEST.py` and `nodeDiscoverySubs_TEST.py` under `python/test` directory. Each one shall be run in a separate terminal before and after modifications to see the results. Before modification, it is not possible to receive the messages by the subscribers due to the issue I mentioned. However, after my modification, the subscribers are able to receive their messages. 
Finally, the proposed modification can also be verified from `CLI`. When the publisher TEST is run, it prints the command that can be run in a shell to set the `DISCOVERY` vars to listen to their publications.

NOTE: The following tests failed:
`NodeTest.ScopeProcess`  and  `NodeTest.ScopeProcessRemap`, because the test runs by creating publisher and subscriber in different threads. Due to the proposed modification, these 2 tests will fail.

## Checklist
- [X] Added tests
- [X] Updated documentation (as needed)
- [X] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

### DIFF
--- a/src/NodeShared.cc
+++ b/src/NodeShared.cc
@@ -159,11 +159,18 @@ NodeShared *NodeShared::Instance()
   // is not shared between different processes.
 
   static std::shared_mutex mutex;
-  static std::unordered_map<unsigned int, NodeShared*> nodeSharedMap;
+  static std::unordered_map<std::string, NodeShared*> nodeSharedMap;
 
   // Get current process ID.
   auto pid = getProcessId();
-
+  // Get current thread ID, so we can create multiple NodeShared per Process, per thread
+  auto tid = std::this_thread::get_id();
+  // Create a unique id using both pid and tid
+  std::stringstream ss;
+  ss << tid;
+  std::stringstream unique_ss;
+  unique_ss << std::hex << pid << "-" << ss.str();
+  auto uid = unique_ss.str();
   // Check if there's already a NodeShared instance for this process.
   // Use a shared_lock so multiple threads can read simultaneously.
   // This will only block if there's another thread locking exclusively
@@ -173,7 +180,7 @@ NodeShared *NodeShared::Instance()
   try
   {
     std::shared_lock readLock(mutex);
-    return nodeSharedMap.at(pid);
+    return nodeSharedMap.at(uid);
   }
   catch (...)
   {
@@ -182,7 +189,7 @@ NodeShared *NodeShared::Instance()
     // not an already constructed NodeShared instance for this process.
     std::lock_guard writeLock(mutex);
 
-    auto iter = nodeSharedMap.find(pid);
+    auto iter = nodeSharedMap.find(uid);
     if (iter != nodeSharedMap.end())
     {
       // There's already an instance for this process, return it.
@@ -190,7 +197,7 @@ NodeShared *NodeShared::Instance()
     }
 
     // No instance, construct a new one.
-    auto ret = nodeSharedMap.insert({pid, new NodeShared});
+    auto ret = nodeSharedMap.insert({uid, new NodeShared});
     assert(ret.second);  // Insert operation should be successful.
     return ret.first->second;
   }
--- a/src/NodeShared.cc
+++ b/src/NodeShared.cc
@@ -163,7 +163,8 @@ NodeShared *NodeShared::Instance()
 
   // Get current process ID.
   auto pid = getProcessId();
-  // Get current thread ID, so we can create multiple NodeShared per Process, per thread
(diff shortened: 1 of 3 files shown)


src/NodeShared.cc
```diff
@@ -159,11 +159,19 @@ NodeShared *NodeShared::Instance()
   // is not shared between different processes.
 
   static std::shared_mutex mutex;
-  static std::unordered_map<unsigned int, NodeShared*> nodeSharedMap;
+  static std::unordered_map<std::string, NodeShared*> nodeSharedMap;
 
   // Get current process ID.
   auto pid = getProcessId();
-
+  // Get current thread ID, so we can create multiple NodeShared per Process
```
azeey, line 166: I believe this will mean nodes in the same process but in different threads will be treated as remote nodes, which has significant performance implications; messages would have to be serialized and sent over the network instead of being delivered via a direct function call. If this is the case, I don't think we can accept this change. @caguero, can you verify?

azeey (changes requested):

src/NodeShared.cc
AmalDevHaridevan: I see. Do you think a workaround using NodeOptions will work. For example, when instantiating with default NodeOptions, transport exhibits the current behaviour. However, by passing an additional argument such as thread_unique in NodeOptions, which will retrieve a thread unique NodeShared instead of process unique NodeShared.

azeey: Yeah, that could work.

AmalDevHaridevan: Ok Cool!

AmalDevHaridevan:
Hi Azeey,

I found that there is an alternate approach to my use case. I feel like implementing my solution requires a lot of modifications, and after testing it, even though it works, it feels like I modified some fundamental architecture. Firstly, I do not know why the architecture was used in the first place, so modifying it may come with consequences.
So I decided to settle with an IPC solution to my problem (even though its much slower).

PR #565, opened 2025-01-09
Unhide cmake warning when Python3 Development is not found

scpeters:
# 🦟 Bug fix

Unhides a cmake warning when python bindings will not be built

## Summary

A cmake warning should be printed when python bindings will not be built (controlled by the logic around `add_subdirectory(python)`), but the warning was in a separate part of the code, allowing the warning to be missed if the `Python3_Development_FOUND` variable changed in between.

Since `GZ_BUILD_WARNING("")` only generates a cmake warning if invoked before `gz_configure_build`, I switched to use `message(WARNING "")` in https://github.com/gazebosim/gz-transport/pull/565/commits/e464330418197347b4ce208be354864405cf5779.

Also, the failed Ubuntu Noble CI build is expected because finding python is currently broken for all platforms, but only our Ubuntu Noble CI builds report cmake warnings. I believe finding python should be fixed once https://github.com/gazebosim/gz-msgs/pull/479 is merged and released. I think it would be worth merging this so that our broken state is accurately reflected.

## Checklist
- [ ] Added tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,7 @@ if (NOT SKIP_PYBIND11)
   if (Python3_Development_FOUND)
     add_subdirectory(python)
   else ()
-    GZ_BUILD_WARNING("Python development libraries are missing: Python interfaces are disabled.")
+    message(WARNING "Python development libraries are missing: Python interfaces are disabled.")
   endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


scpeters:
this should be ported to gz-math and sdformat as well, since they use the same logic pattern

caguero (approved):

scpeters:
https://github.com/Mergifyio backport gz-transport13

PR #601, opened 2025-04-09
Zenoh - Part 2: Configure zeromq/zenoh and pub sub

caguero:
# 🎉 New feature

* It allows to use `zeromq` or `zenoh` as the underlying middleware. You can use the environment variable `GZ_TRANSPORT_IMPLEMENTATION` with the values `zeromq` or `zenoh`.
* Adds pub sub to the zenoh branch.

How to test the pub/sub?

Compile the examples and try pub/sub with:

```
GZ_TRANSPORT_IMPLEMENTATION=zeromq ./publisher
```
and
```
GZ_TRANSPORT_IMPLEMENTATION=zeromq ./subscriber
```
You can also try the generic subscriber:
```
GZ_TRANSPORT_IMPLEMENTATION=zeromq ./subscriber_generic
```
And the raw subscriber:
```
GZ_TRANSPORT_IMPLEMENTATION=zeromq ./subscriber_raw
```

You can repeat the previous examples replacing `zeromq` with `zenoh`.

All the tests should pass when using `zeromq` but not with `zenoh` as there are more pieces to be merged.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #621, opened 2025-05-02
Add Ubuntu Noble - Zenoh github action

iche033:
# 🎉 New feature

## Summary

Add a new Ubuntu github action for running gz-transport tests with zenoh. You should see a new check in this PR: `[Ubuntu CI / Ubuntu Noble CI - Zenoh (pull_request)`

This introduces a `GZ_TRANSPORT_DEFAULT_IMPLEMENTATION` cmake arg that can be configured at compile time to set the default transport implementation to use. By default it's `zeromq`. This cmake arg is set by the new github action in `.github/workflow/ci.yml` to `zenoh`. The `GZ_TRANSPORT_IMPLEMENTATION` env var should continue to work the same way, and when set, it'll override `GZ_TRANSPORT_DEFAULT_IMPLEMENTATION`.

Note: in the long run, we should consider implementing a common test frame work similar to the one in [gz-rendering/test/common](https://github.com/gazebosim/gz-rendering/tree/gz-rendering9/test/common_test) and [gz-physics/test/common](https://github.com/gazebosim/gz-physics/tree/gz-physics8/test/common_test) so we can run all tests with different transport implementations one after another in the same job.

## Checklist
- [ ] Added tests


caguero (commented):
Is it expected to fail?

caguero:
> > Is it expected to fail?
> 
> There are test failures in the build, which I also want to check with you. The services are not supported yet in this branch (branched off of #601) so I assume that's why the tests are failing. I also tried running `UNIT_Node_TEST` locally with zenoh but it crashes like in the CI job. So I think it's expected that not all tests should pass right?

Yes, not all tests should pass.

iche033:
> > > Is it expected to fail?
> > 
> > 
> > There are test failures in the build, which I also want to check with you. The services are not supported yet in this branch (branched off of #601) so I assume that's why the tests are failing. I also tried running `UNIT_Node_TEST` locally with zenoh but it crashes like in the CI job. So I think it's expected that not all tests should pass right?
> 
> Yes, not all tests should pass. For some reason I wasn't able to see any test result. That's why I was confused.
> 
> Could we limit this job to just a test (for now) that we know it passes?

yes disabled all tests that are failing / crashing / timing out in zenoh. [77837f0](https://github.com/gazebosim/gz-transport/pull/621/commits/77837f06fc732b90fbd6e271d3c29e006d444ba7). Note I added a `getTransportImplementation()` free function so that we can access it in python tests to determine whether or not to disable them. CI should be green now

caguero (approved):

PR #628, opened 2025-05-09
Migrate bazel build setup to use bzlmod

shameekganguly:
# 🎉 New feature

## Summary
Bazel is migrating away from the WORKSPACE setup for dependency management, which is replaced by bzlmod (https://bazel.build/external/migration).

This change makes bazel build for gz-transport compatible with bzlmod, and is similar to https://github.com/gazebosim/sdformat/pull/1493 which added bzlmod support for sdformat.

A couple notes:

- Most of the bazel tests had to be disabled since they depend on gz-common/testing which is not yet migrated to bzlmod. I will re-enable these tests once gz-common/testing is migrated to bzlmod.
- I had to push `BatchPrivate` into the versioned namespace to match the forward declaration in https://github.com/gazebosim/gz-transport/blob/gz-transport14/log/include/gz/transport/log/Batch.hh to compile with the bzlmod setup.

## Test it
```shell
$ bazel build :all
$ bazel test :all
```

## Checklist
- [ ] Added tests


BUILD.bazel
```diff
@@ -1,38 +1,39 @@
...
 )
 
-licenses(["notice"])  # Apache-2.0
+licenses(["notice"])
 
-exports_files(["LICENSE"])
+exports_files([
+    "LICENSE",
+    "MODULE.bazel",
+    "package.xml",
```
iche033, line 22: not sure if this is needed or not but just noticed that `package.xml` is not in the `exports_files` list in gz-msgs PR https://github.com/gazebosim/gz-msgs/pull/507

iche033 (approved):
looks good to me

BUILD.bazel
shameekganguly: Yes, this is needed here since `package.xml` is referenced in the `//log:Config` target in log/BUILD.bazel

PR #630, opened 2025-05-12
Zenoh - Part 3: Graph cache pub/sub

caguero:
Adds the first pass of a graph cache working with Zenoh. It only supports pub/sub.

### How to test it

Compile the examples and try publisher/subscriber with:

```
GZ_TRANSPORT_IMPLEMENTATION=zenoh ./publisher
```
and
```
GZ_TRANSPORT_IMPLEMENTATION=zenoh ./subscriber
```
and
```
GZ_TRANSPORT_IMPLEMENTATION=zenoh gz topic -l
```
You should see the topic `/foo`.

You can repeat the previous examples replacing `zenoh` with `zeromq` to verify that nothing is broken.

## Checklist
- [ ] Added tests

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.**

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -10,7 +10,7 @@ release will remove the deprecated code.
 
 ### Removed
 
-1. The `Discovery.hh` file is not longer installed.
+1. The `Discovery.hh` file is no longer installed.
     * [GitHub pull request 630](https://github.com/gazebosim/gz-transport/pull/630)
 
 ### Breaking Changes


src/NodeShared.cc
```diff
@@ -297,32 +312,25 @@ NodeShared::NodeShared()
...
-  // Start the discovery services.
-  if (this->GzImplementation() == "zeromq")
-  {
+    // Start the discovery services.
     this->dataPtr->msgDiscovery->Start();
     this->dataPtr->srvDiscovery->Start();
   }
+  else if (this->GzImplementation() == "zenoh")
+  {
+    this->dataPtr->msgDiscovery->Start(this->Session());
```
iche033, line 332: I think you need to guard this with `#ifdef HAVE_ZENOH` because `Session()` function is only available if zenoh is defined.

iche033: I found this in my local build with no zenoh deps installed. Maybe we should consider setting up another gz-transport github action job with no zenoh deps installed.

caguero: Thanks! Added in [1532b7f](https://github.com/gazebosim/gz-transport/pull/630/commits/1532b7f92a68a70742e66754b8840b69be60f4ff).

include/gz/transport/SubscriptionHandler.hh
```diff
@@ -67,9 +67,11 @@ namespace gz
     class GZ_TRANSPORT_VISIBLE SubscriptionHandlerBase
     {
       /// \brief Constructor.
+      /// \param[in] _pUuid UUID of the process registering the handler.
       /// \param[in] _nUuid UUID of the node registering the handler.
       /// \param[in] _opts Subscription options.
       public: explicit SubscriptionHandlerBase(
+        const std::string &_pUuid,
```
iche033, line 74: given that this changes the constructor signature in a public header that's installed, we may want to document this in the migration guide.

src/Discovery.hh
iche033, line 1: will this break users code because this header is no longer installed? Document this in migration guide?

iche033 (commented):
added some comments about documenting breaking changes, otherwise looks good

src/Discovery.hh
caguero: Yes but I think it's very unlike. This is a very low level internal class and nobody should really use it. I documented it in [79dc05d](https://github.com/gazebosim/gz-transport/pull/630/commits/79dc05d3d764a024da1d9766fa977260b3b47957).

Migration.md
```diff
@@ -8,8 +8,23 @@ release will remove the deprecated code.
 
 ## Gazebo Transport 14.X to 15.X
 
+### Removed
+
+1. The `Discovery.hh` file is not longer installed.
```
iche033, line 13:
```suggestion
1. The `Discovery.hh` file is no longer installed.
```

PR #657, opened 2025-06-11
Zenoh - Part 5: Implement graph cache for async services

caguero:
Adds Zenoh graph cache support for async services.

### How to test it

Compile the examples and run a service provider:

```
GZ_TRANSPORT_IMPLEMENTATION=zenoh ./responser
```
and
```
GZ_TRANSPORT_IMPLEMENTATION=zenoh gz service -l
```
You should see the service `/echo`.

You can repeat the previous examples replacing `zenoh` with `zeromq` to verify that nothing is broken.

## Checklist
- [ ] Added tests

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.**

### DIFF
--- a/include/gz/transport/RepHandler.hh
+++ b/include/gz/transport/RepHandler.hh
@@ -61,7 +61,10 @@ namespace gz::transport
   class GZ_TRANSPORT_VISIBLE IRepHandler
   {
     /// \brief Constructor.
-    public: IRepHandler();
+    /// \param[in] _pUuid Process UUID.
+    /// \param[in] _nUuid Node UUID.
+    public: explicit IRepHandler(const std::string &_pUuid,
+                                 const std::string &_nUuid);
 
     /// \brief Destructor.
     public: virtual ~IRepHandler();
@@ -126,7 +129,7 @@ namespace gz::transport
     : public IRepHandler
   {
     // Documentation inherited.
-    public: RepHandler() = default;
+    using IRepHandler::IRepHandler;
 
     /// \brief Set the callback for this handler.
     /// \param[in] _cb The callback with the following parameters:
--- a/include/gz/transport/TopicUtils.hh
+++ b/include/gz/transport/TopicUtils.hh
@@ -247,7 +247,7 @@ namespace gz::transport
     /// qualified topic name.
     public: static const uint16_t kMaxNameLength = 65535;
 
-    /// \param The separator used within the liveliness token.
+    /// \brief The separator used within the liveliness token.
     public: static constexpr const char *kTokenSeparator = "@";
 
     /// \brief A common prefix for all liveliness tokens.
--- a/include/gz/transport/TopicUtils.hh
+++ b/include/gz/transport/TopicUtils.hh
@@ -164,6 +164,76 @@ namespace gz::transport
       std::string &_entityType,
       std::string &_msgType);
 
+    /// \brief Decompose a Zenoh liveliness token into its components.
+    ///
+    /// Given a Zenoh liveliness token with the following syntax:
+    ///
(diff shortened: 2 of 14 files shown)


include/gz/transport/TopicUtils.hh
```diff
@@ -164,6 +164,76 @@ namespace gz::transport
       std::string &_entityType,
       std::string &_msgType);
 
+    /// \brief Decompose a Zenoh liveliness token into its components.
+    ///
+    /// Given a Zenoh liveliness token with the following syntax:
+    ///
+    /// \<PREFIX\>\@\<PARTITION\>\@\<NAMESPACE\>/\<TOPIC\>@\<ProcUUID\>
+    /// \@\<NodeUUID\>\@\<EntityType\>\@\<ReqType\>\@\<RepType\>
```
iche033, line 172: I see some differences with the ROS 2 graph cache liveliness token [design](https://github.com/ros2/rmw_zenoh/blob/rolling/docs/design.md#graph-cache) so I have a couple of questions.

1. Naively, is it worth (or feasible) using the same format? e.g. work with the ROS 2 team on the possibility of reusing code from [liveliness_utils](https://github.com/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/src/detail/liveliness_utils.hpp) to construct the token.
2. If not, do you imagine that we'll need to have a way to convert between liveliness token formats at runtime if we want to enable cross-talk with ROS 2 on the zenoh level?

include/gz/transport/RepHandler.hh
```diff
@@ -61,7 +61,10 @@ namespace gz::transport
   class GZ_TRANSPORT_VISIBLE IRepHandler
   {
     /// \brief Constructor.
-    public: IRepHandler();
+    /// \param[in] _pUuid Process UUID.
+    /// \param[in] _nUuid Node UUID.
+    public: explicit IRepHandler(const std::string &_pUuid,
+                                 const std::string &_nUuid);
```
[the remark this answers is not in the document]
caguero: Added in [27278e6](https://github.com/gazebosim/gz-transport/pull/657/commits/27278e6ea7ce07d4ca68ee7617ad225ded3035b1).

include/gz/transport/TopicUtils.hh
caguero: I think we're not that far but not sure if both projects have the ability to agree in the convention without changes. How about this proposal? (1) We leave the code with our convention for now, (2) in parallel I'll talk with @yadunund to better understand the ROS fields of that convention, (3) and when we have a complete version ready, we'll work on the necessary changes to either get both projects closer or reuse the same token and get fully aligned.

iche033: the proposal sounds good to me. We should have a follow-up discussion on some convergence strategy, ideally before the Jetty release

iche033 (approved):

include/gz/transport/TopicUtils.hh
Yadunund: Happy to align on conventions~
We can even move headers currently in `rmw_zenoh_cpp/src` into a common library that both projects can leverage.

Would be also great to discuss how we can interoperate between gz-transport and rmw_zenoh. If we can [encode](https://zenoh-cpp.readthedocs.io/en/1.4.0/publish_subscribe.html#_CPPv4N5zenoh9Publisher10PutOptions8encodingE) information on the serialization format, have type support for protobuf in ROS, and have some TypeAdapters defined between common [gz-msgs](https://github.com/gazebosim/gz-msgs/tree/gz-msgs11/proto/gz/msgs) and ROS types, we could have ROS nodes directly publish/subscribe to GZ nodes. (skip the ros_gz_bridge entirely)

PR #665, opened 2025-07-15
Integrate Zenoh

caguero:
# 🎉 New feature

## Summary

This PR merges all functionality that we've been merging into the `zenoh_main` branch back into `main`.

## Checklist
- [x] Added tests

### DIFF
--- a/src/Node.cc
+++ b/src/Node.cc
@@ -1145,7 +1145,7 @@ Node::Publisher Node::Advertise(const std::string &_topic,
   auto currentTopics = this->AdvertisedTopics();
 
   if (std::find(currentTopics.begin(), currentTopics.end(),
-        fullyQualifiedTopic) != currentTopics.end())
+        _topic) != currentTopics.end())
   {
     std::cerr << "Topic [" << topic << "] already advertised. You cannot"
       << " advertise the same topic twice on the same node."
--- a/src/Discovery.hh
+++ b/src/Discovery.hh
@@ -324,7 +324,8 @@ namespace gz
             if (this->verbose)
             {
               std::cout << ">> [LivelinessSubscriber] New alive token ('"
-                        << _sample.get_keyexpr().as_string_view() << "')\n";
+                        << _sample.get_keyexpr().as_string_view() << "')"
+                        << std::endl;
             }
           }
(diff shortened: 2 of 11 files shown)


iche033:
> I'll look into the bazel failures.

bazel builds should be fixed by #666

src/Node.cc
```diff
@@ -526,25 +566,44 @@ bool Node::Publisher::Publish(const ProtoMsg &_msg)
...
+  else if (impl == "zenoh")
   {
-    delete[] msgBuffer;
+    if (subscribers.haveRemote)
+    {
+      zenoh::Publisher::PutOptions options;
+      // Add message type as an attachment.
+      options.attachment = this->dataPtr->publisher.MsgTypeName();
+
+      this->dataPtr->zPub->put(msgBuffer, std::move(options));
```
azeey, line 601: I was testing this with gz-sim and ran into error messages that said 

```
SubscriptionHandler::CreateMsg() error: ParseFromString failed
```

After investigating a bit, I pinpointed the problem to this line. What I believe is happening here is that `zenoh::Bytes` copies `msgBuffer` as though it was. a c-string, so if a null character happens to be in the message, the resulting message published would be truncated. The solution here is to use the constructor that takes a size argument which takes ownership of the buffer instead of copying it and relies on the provided size, so null characters are not a problem. After making this change and my suggested change above, I was able to run `gz sim shapes.sdf` succesfully 🎉 


```suggestion
      // Zenoh will call this lambda once Bytes objects are destroyed
      auto deleter = [](uint8_t *_buffer) { delete[] _buffer; };
      auto zMsgBuffer = reinterpret_cast<uint8_t *>(msgBuffer);
      this->dataPtr->zPub->put(zenoh::Bytes(zMsgBuffer, msgSize, deleter),
                               std::move(options));

```

azeey (changes requested):

.github/ci/packages-noble.apt
```diff
@@ -1 +1,3 @@
 cppzmq-dev
+ros-jazzy-ros-environment
```
arjo129, line 2: Just noticed this. Does this mean that when we build gazebo without ros support we should pass the `-DGZ_TRANSPORT_ENABLE_ZENOH=OFF` ?

PR #681, opened 2025-08-13
Fix log component tests with Zenoh

azeey:
# 🦟 Bug fix

Builds on top of #680 

## Summary
Fixes most of the test failures in the `log` component when ran with Zenoh.

TODO:
- [x] `twoProcPubSub.TopicInfo`: `Node::TopicInfo` returns more publishers than expected. I'm guessing we're not deleting publishers effectively when they go away.
- [x] `twoProcPubSub.PubSubTwoProcsScopedPub`:  `Publisher::HasConnections` is returning an unexpected true. This might be a wrong assumption in the test that it takes a certain amount of time before subscribers are seen on the network. https://github.com/azeey/gz-transport/blob/2d85e40bb2ea258f35188ef93060cd98df0ce8b0/test/integration/twoProcsPubSub.cc#L436-L438
- [ ] `recorder.BeginRecordingPatternBeforeAdvertisement` and `recorder.BeginRecordingAllBeforeAdvertisement`: The `Recorder` uses the `Discovery` class which now requires starting the Zenoh-specific `Discovery::Start` function. Before I made any changes, the `OnAdvertisement` callback was not being called when new publishers were available on the network. I tried to fix by calling the new `Discovery::Start` function in `Recorder`. This by itself did not work. I then tried adding the missing invokation of `connectionCb`. This however causes a crash in `libzenohc.so`. However, it sometimes works under `gdb`, so it seems promising.

## Checklist
- [ ] Added tests

### DIFF
--- a/log/test/integration/playback.cc
+++ b/log/test/integration/playback.cc
@@ -202,6 +202,8 @@ TEST(playback, GZ_UTILS_TEST_DISABLED_ON_MAC(ReplayNoSuchTopic))
 /// the original.
 TEST(playback, GZ_UTILS_TEST_DISABLED_ON_MAC(ReplayLogRegex))
 {
+  CHECK_UNSUPPORTED_IMPLEMENTATION("zenoh")
+
   std::vector<std::string> topics = {"/foo", "/bar", "/baz"};
 
   std::vector<MessageInformation> incomingData;
@@ -269,6 +271,8 @@ TEST(playback, GZ_UTILS_TEST_DISABLED_ON_MAC(ReplayLogRegex))
 /// that the playback matches the original.
 TEST(playback, GZ_UTILS_TEST_DISABLED_ON_MAC(RemoveTopic))
 {
+  CHECK_UNSUPPORTED_IMPLEMENTATION("zenoh")
+
   std::vector<std::string> topics = {"/foo", "/bar", "/baz"};
 
   std::vector<MessageInformation> incomingData;


log/src/Recorder.cc
```diff
@@ -200,6 +200,9 @@ Recorder::Implementation::Implementation()
 
   this->discovery->ConnectionsCb(cb);
   this->discovery->Start();
+  this->discovery->Start(shared->Session(),
```
caguero, line 203: We need to check `GzImplementation()` here and choose the right `Start()` method. We'll need to add and `#ifdef` in the `zenoh` else branch.

log/src/Recorder.cc
```diff
@@ -310,6 +313,7 @@ int64_t Recorder::Implementation::AddTopic(const std::regex &_pattern)
   this->node.TopicList(allTopics);
   for (auto topic : allTopics)
   {
+    std::cout << "Checking: " << topic << "\n";
```
caguero, line 316: Remove?

log/src/Recorder.cc
azeey: https://github.com/gazebosim/gz-transport/pull/681/commits/6e7a8d4324fb7a968cfcf95ee4f6a3771e5c0e7e

caguero (commented):
`INTEGRATION_playback` seems funky.

caguero (approved):

PR #731, opened 2025-10-03
Only check GZ_TRANSPORT_TOPIC_STATISTICS once

### LINKED ISSUE #701: Occasional crash on startup in discovery
## Environment
* OS Version: ubuntu 24.04
* Source or binary build?
  source build, `gz-transport15` branch
  

## Description
* Expected behavior: gazebo + gz-transport starts up fine without crashing
* Actual behavior: Once in a while I get a crash with a backtrace pointing to gz-transport

## Steps to reproduce

1. launch gz-sim with any world

## Output

Crash happens occasionally, and gz-sim spits out the backtrace below that points to the use of `std::getenv`

<details><summary>backtrace</summary>

```
#7    Source "../../../../../src/libstdc++-v3/src/c++11/thread.cc", line 104, in execute_native_thread_routine [0x7f1516a76db3]
#6    Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/Discovery.hh", line 1032, in RecvMessages [0x7f15177bf1a1]
       1030:           if (pollSockets(this->sockets, timeout))
       1031:           {
      >1032:             this->RecvDiscoveryUpdate();
       1033: 
       1034:             if (this->verbose)
       1035:               this->PrintCurrentState();
#5    Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/Discovery.hh", line 1100, in __builtin_strlen [0x7f15177bf067]
       1097:                 << srcAddr << ": " << srcPort << std::endl;
       1098:             }
       1099: 
      >1100:             this->DispatchDiscoveryMsg(srcAddr, rcvStr + sizeof(len), len);
       1101:           }
       1102:         }
       1103:         else if (received < 0)
#4  | Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/Discovery.hh", line 1266, in operator()
    |  1264:           {
    |  1265:             if (subscribersReqCb)
    | >1266:               subscribersReqCb();
    |  1267: 
    |  1268:             break;
      Source "/usr/include/c++/13/bits/std_function.h", line 591, in DispatchDiscoveryMsg [0x7f15177be7cc]
        588:       {
        589:    if (_M_empty())
        590:      __throw_bad_function_call();
      > 591:    return _M_invoker(_M_functor, std::forward<_ArgTypes>(__args)...);
        592:       }
        593: 
        594: #if __cpp_rtti
#3  | Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/NodeShared.cc", line 1390, in SendSubscribersRep
    |  1388:   // Reply to the SUBSCRIBERS_REQ with multiple SUBSCRIBERS_REP.
    |  1389:   for (auto const &publisher : pubs)
    | >1390:     this->dataPtr->msgDiscovery->SendSubscribersRep(publisher);
    |  1391: }
      Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/Discovery.hh", line 565, in OnSubscribers [0x7f15177a8462]
        562:       /// \param[in] _pub Information to send.
        563:       public: void SendSubscribersRep(const MessagePublisher &_pub) const
        564:       {
      > 565:         this->SendMsg(
        566:           DestinationType::ALL, msgs::Discovery::SUBSCRIBERS_REP, _pub);
        567:       }
#2  | Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/Discovery.hh", line 1382, in Version
    |  1380:       {
    |  1381:         gz::msgs::Discovery discoveryMsg;
    | >1382:         discoveryMsg.set_version(this->Version());
    |  1383:         discoveryMsg.set_type(_type);
    |  1384:         discoveryMsg.set_process_uuid(this->pUuid);
      Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/Discovery.hh", line 1574, in SendMsg<gz::transport::v15::MessagePublisher> [0x7f151778cd55]
       1571:         static std::string gzStats;
       1572:         static int topicStats;
       1573: 
      >1574:         if (env("GZ_TRANSPORT_TOPIC_STATISTICS", gzStats) && !gzStats.empty())
       1575:         {
       1576:           topicStats = (gzStats == "1");
       1577:         }
#1    Source "/home/iche/code/gz_j_ws_noble/src/gz-transport/src/Helpers.cc", line 41, in env [0x7f1517777144]
         38:     size_t sz = 0;
         39:     _dupenv_s(&v, &sz, _name.c_str());
         40: #else
      >  41:     v = std::getenv(_name.c_str());
         42: #endif
         43:     if (v)
         44:     {
#0    Source "./stdlib/getenv.c", line 31, in getenv [0x7f15167c0806]
Segmentation fault (Address not mapped to object [0x5570fb534])
Segmentation fault (core dumped)
``

</details>

### PR DESCRIPTION
**Pull request by caguero**
# 🦟 Bug fix

Fixes #701 

## Summary

According to #701, checking an environment variable every time that we're inside the `Version()` function might cause a crash. This patch only checks the environment variable once.

There's some behavior change here because you cannot change the value of the environment variable in the middle of a gz-transport session anymore. I think it's probably a good idea not allowing that but I can be convinced otherwise.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/Discovery.hh
+++ b/src/Discovery.hh
@@ -1570,10 +1570,13 @@ namespace gz
       {
         static std::string gzStats;
         static int topicStats;
+        static bool versionInitialized = false;
 
-        if (env("GZ_TRANSPORT_TOPIC_STATISTICS", gzStats) && !gzStats.empty())
+        if (!versionInitialized &&
+            env("GZ_TRANSPORT_TOPIC_STATISTICS", gzStats) && !gzStats.empty())
         {
           topicStats = (gzStats == "1");
+          versionInitialized = true;
         }
 
         return this->kWireVersion + (topicStats * 100);

### CODE REVIEW

File: src/Discovery.hh [hunk 1]
```diff
@@ -1570,10 +1570,13 @@ namespace gz
       {
         static std::string gzStats;
```
azeey, line 1571: We can make `gzStats` a non-static std::string then. It's always good to remove static variables with non-trivial destructors https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

src/Discovery.hh
```diff
@@ -1570,10 +1570,13 @@ namespace gz
...
         static int topicStats;
+        static bool versionInitialized = false;
```
azeey, line 1573: Should these be `std::atomic` in case they are accessed from mulitple threads?

src/Discovery.hh
```diff
@@ -1570,10 +1570,13 @@ namespace gz
...
 
-        if (env("GZ_TRANSPORT_TOPIC_STATISTICS", gzStats) && !gzStats.empty())
+        if (!versionInitialized &&
+            env("GZ_TRANSPORT_TOPIC_STATISTICS", gzStats) && !gzStats.empty())
         {
           topicStats = (gzStats == "1");
+          versionInitialized = true;
```
azeey, line 1579: If `GZ_TRANSPORT_TOPIC_STATISTICS` is not set, this `if` block will not get executed, so we'll end up checking the environment variable everytime still. How about putting this `versionInitialized = true` after the `if` block?

azeey (commented):
I would agree that environment variables should not be used to change behavior after the program has started. If we want to do that, we should be providing a C++ API for it.

src/Discovery.hh
caguero: Good idea. [40fbf54](https://github.com/gazebosim/gz-transport/pull/731/commits/40fbf5405a32be2857ab525572dab02fa427f9d5)

src/Discovery.hh
caguero: Changed in [40fbf54](https://github.com/gazebosim/gz-transport/pull/731/commits/40fbf5405a32be2857ab525572dab02fa427f9d5)

src/Discovery.hh
```diff
@@ -1568,14 +1568,18 @@ namespace gz
...
 
-        if (env("GZ_TRANSPORT_TOPIC_STATISTICS", gzStats) && !gzStats.empty())
+        if (!versionInitialized &&
+            env("GZ_TRANSPORT_TOPIC_STATISTICS", gzStats) && !gzStats.empty())
         {
           topicStats = (gzStats == "1");
         }
 
+        versionInitialized = true;
+
```
azeey, line 1582: hmm, thinking about it some more. It's probably still not safe from a data race. `versionInitialized`, even though atomic, the whole processes of checking the environment variable and setting `topicStats` is not atomic. 

My suggestion would be to use [`std::call_once`](https://en.cppreference.com/w/cpp/thread/call_once.html) with a local lambda

azeey: Okay, but note that we'll need to backport this because https://github.com/gazebosim/gz-transport/issues/701 occurs on Jetty.

PR #801, opened 2026-02-28
Pass std::function by value in Node::Advertise and Node::Request

Ryujiyasu:
# 🎉 New feature

Closes #203

## Summary

Applies the same approach as #382 (`Subscribe`) to the `Advertise` and `Request` methods, allowing users to pass capturing lambda callbacks directly without wrapping them in a `std::function` variable first.

- Changed `std::function` parameters from pass-by-reference to pass-by-value in `Node::Advertise` (2 overloads) and `Node::Request` (2 overloads)
- Used `std::move` to forward callbacks efficiently, consistent with #382
- Added 3 unit tests that pass capturing lambdas directly to `Advertise` and `Request`

## Test it

```bash
cd build
./bin/UNIT_Node_TEST --gtest_filter="*DirectLambda*"
```

Three new tests verify capturing lambdas work directly:
- `ServiceCallAsyncDirectLambda`
- `ServiceCallWithoutInputAsyncDirectLambda`
- `ServiceCallWithoutOutputAsyncDirectLambda`

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [x] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Generated-by: Claude Opus 4.6

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

PR #828, opened 2026-03-11
Revert #801

caguero:
# 🦟 Bug fix

Unfortunately, this is a rollback of #801.  

PR #801 changed `std::function` parameters from pass-by-reference to pass-by-value in `Node::Advertise` and `Node::Request`.  This causes overload ambiguity when a non-capturing lambda is passed with explicit template arguments: both the function pointer overload and the `std::function` overload become equally viable candidates (each requiring a user-defined conversion), and the compiler cannot choose between them.                                                              
                                                                                                                           
 The issue was found compiling gz-sim:                                                                                  

```
  error: call of overloaded 'Request(...)' is ambiguous
    this->node.Request<msgs::StringMsg, msgs::Boolean>(
        "/world/" + worldName + "/level/set_performer", req,
        [](const msgs::Boolean &, const bool) {});
```

This patch reverts the pass-by-value change from #801 and adds some regression tests with non-capturing lambdas. There's a solution for this problem preserving #801 but I'm not convinced it's worth the extra set of functions needed.

We can go back to it if we can come up with a good solution.

CC/ @Ryujiyasu

## Checklist
- [x] Added tests


src/Node_TEST.cc
```diff
@@ -1674,107 +1674,71 @@ TEST(NodeTest, ServiceCallWithoutOutputAsyncLambda)
...
-  // Capturing lambdas cannot convert to function pointers, so this tests
-  // the std::function pass-by-value overload.
+  // Advertise with a non-capturing lambda and explicit template args.
   EXPECT_TRUE((node.Advertise<msgs::Int32, msgs::Int32>(g_topic,
-    [&advExecuted](const msgs::Int32 &_req, msgs::Int32 &_rep) -> bool
+    [](const msgs::Int32 &_req, msgs::Int32 &_rep) -> bool
     {
-      EXPECT_EQ(_req.data(), data);
       _rep.set_data(_req.data());
-      advExecuted = true;
```
azeey, line 1696: If we remove `advExecuted`, the test would succeed even if the callback was never executed. Should we have one callback that is capturing and one that is not?

azeey (commented):
It's unfortunate that we have to rollback. I was looking forward to using lambdas directly as callbacks instead of going through std::function. The typical way this is done in the STL is to take a template for the type of the callback function (see https://en.cppreference.com/w/cpp/utility/functional/invoke.html)

caguero:
> It's unfortunate that we have to rollback. I was looking forward to using lambdas directly as callbacks instead of going through std::function. The typical way this is done in the STL is to take a template for the type of the callback function (see https://en.cppreference.com/w/cpp/utility/functional/invoke.html)

[This](https://github.com/gazebosim/gz-transport/compare/main...fix-advertise-ambiguity-v2?expand=1) is an alternative solution to revert #801 that solves the overload ambiguity.

@azeey , what do you think?

azeey:
I'd say let's merge the revert to unblock downstream packages and continue the discussion about alternatives on a new PR.

azeey (approved):

azeey:
It would be best if the revert was just a clean revert to keep a clear history.

caguero:
> It would be best if the revert was just a clean revert to keep a clear history.

Done, it's a direct revert of #801, now.

azeey (approved):

azeey:
Please "Rebase and Merge"

PR #831, opened 2026-03-16
Use rotary unversioned packages for CI

j-rivero:
Part of https://github.com/gazebo-tooling/release-tools/issues/1446

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -1,8 +1,8 @@
-libgz-cmake-dev
-libgz-math-dev
-libgz-msgs-dev
-libgz-tools-dev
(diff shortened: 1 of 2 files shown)


.github/ci/packages.apt
```diff
@@ -13,5 +13,5 @@ python3-dev
 python3-psutil
 python3-pybind11
 python3-pytest
-python3-gz-msgs12
+gz-rotary-msgs-python
```
azeey, line 16: Is this right? I would have expected `python3-gz-rotary-msgs`

j-rivero:
Deprecated in favor of changes using the rotary aliases

PR #842, opened 2026-04-07
Enable Zenoh shared memory

### LINKED ISSUE #754: Zenoh outstanding tasks
* [x] #760 
* [x] #698 
* [x] #867
* [ ] #868 
* [ ] Enable shared memory.

caguero:
# 🎉 New feature

Closes #754

## Summary

This patch integrates Zenoh shared-memory capabilities into gz-transport.

Some benchmarks:

throughput_pubsub

latency_pubsub

svc_twoway_throughput

svc_twoway_latency

## Test it

You can run the benchmarks with:

```
cd example/benchmark
mkdir build && cd build
make
cd ..
python3 run_benchmark.py -b build/bench --quick
```

## Checklist
- [x] Added tests
- [x] Updated migration guide (as needed)


src/ShmHelpers.hh
```diff
@@ -296,63 +289,58 @@ inline namespace GZ_TRANSPORT_VERSION_NAMESPACE
 
   /// \brief Attempt to allocate a writable SHM chunk from the session's
   /// pool. The threshold is checked before touching the provider.
+  /// \param[in] _shm The SHM state of the session.
   /// \param[in] _size Number of bytes to allocate.
   /// \return The chunk, empty if SHM is disabled or not ready, the message
   /// is below threshold, or allocation fails.
-  inline ShmChunk allocShmChunk(std::size_t _size)
+  inline ShmChunk allocShmChunk(ZenohShm &_shm, std::size_t _size)
```
luca-della-vedova: Since all these functions now need access to a `ZenohShm` object why not just make them member functions of `ZenohShm`?

src/ShmHelpers.hh
```diff
@@ -0,0 +1,389 @@
...
+    private: std::optional<zenoh::ZShmMut> buf;
+  };
+
+  /// \brief Attempt to allocate a writable SHM chunk from the session's
+  /// pool. The threshold is checked before touching the provider.
+  /// \param[in] _shm The SHM state of the session.
+  /// \param[in] _size Number of bytes to allocate.
+  /// \return The chunk, empty if SHM is disabled or not ready, the message
+  /// is below threshold, or allocation fails.
+  inline ShmChunk allocShmChunk(ZenohShm &_shm, std::size_t _size)
```
luca-della-vedova: Should this (and below) just be members of the `ZenohShm` object since they need to take a reference to it?

include/gz/transport/detail/Node.hh
```diff
@@ -171,7 +171,10 @@ namespace gz::transport
 #ifdef HAVE_ZENOH
     else if (impl == "zenoh")
     {
-      subscrHandlerPtr->SetCallback(std::move(_cb),
+      // Just store the callback - no per-handler Zenoh subscriber.
+      // The centralized subscriber in NodeShared handles data dispatch.
+      subscrHandlerPtr->SetCallback(std::move(_cb));
+      subscrHandlerPtr->CreateLivelinessToken(
         this->Shared()->Session(), fullyQualifiedTopic);
```
azeey, line 178: Can this change be reverted? The  `SetCallback` that takes the extra arguments now does the same thing as the two function calls here.
 
https://github.com/gazebosim/gz-transport/blob/31a7a5f855e5a3f54357b5fac85e6053bf66b971/include/gz/transport/SubscriptionHandler.hh#L270-L276

src/Node.cc
```diff
@@ -428,24 +444,47 @@ bool Node::Publisher::Publish(const ProtoMsg &_msg)
 #else
   const std::size_t msgSize = static_cast<std::size_t>(_msg.ByteSize());
 #endif
-  char *msgBuffer = nullptr;
 
-  // Only serialize the message if we have a raw subscriber or a remote
-  // subscriber.
-  if (subscribers.haveRaw || subscribers.haveRemote)
+  // Get the transport implementation once for use throughout.
+  std::string impl = this->dataPtr->shared->GzImplementation();
```
azeey: Can we avoid a string copy here? How about storing a bool `this->dataPtr->shared->GzImplementation() == "zenoh"`

src/Node.cc
```diff
@@ -94,6 +95,21 @@ class Node::PublisherPrivate
       zToken(std::make_unique<zenoh::LivelinessToken>(std::move(_zToken)))
   {
   }
+
+  /// \brief Publish string data via SHM if available, falling back to heap.
+  /// Does not take ownership — avoids an intermediate heap allocation when
+  /// the caller already has the data in a std::string (e.g. PublishRaw).
+  public: void PublishViaShmOrHeap(const std::string &_data,
```
azeey, line 102: I noticed this used in only one place. Wondering if it's worth the refactor.

include/gz/transport/SubscriptionHandler.hh
```diff
@@ -164,11 +165,48 @@ namespace gz::transport
...
+      const std::string &_type)>;
+
+    /// \brief Create a protobuf message from a raw buffer.
+    /// Avoids a string copy compared to CreateMsg when the data already
+    /// lives in a contiguous buffer (e.g. Zenoh SHM). Dispatches to the
+    /// deserializer registered with SetMsgFactory(); when none was
+    /// registered (e.g. a handler compiled against older headers), falls
+    /// back to a string copy + the virtual CreateMsg.
+    /// Deliberately non-virtual: a new virtual function would change the
+    /// vtable layout and break ABI within a major release.
```
azeey, line 182: Do we really need to worry about ABI breaks in main? I'd think we should add the virtual function in main and do this workaround in older versions. Alternatively, we can merge this as is so it can be backported, but create a follow up PR that removes the workaround.

src/ReqHandler.cc
```diff
@@ -153,7 +158,16 @@ namespace gz::transport
     std::string payload;
     this->Serialize(payload);
     if (!payload.empty())
-      getOpts.payload = zenoh::Bytes(payload);
+    {
+      // Use SHM for large request payloads to avoid a serialization copy,
+      // using the shared process pool.
+      if (auto shmBytes = makeShmBytes(
```
azeey, line 164: Since we've already serialized the data on the heap (`payload` above), do we need to go through `makeShmBytes`? Can we not simply use `put`/`get` which will automatically use SHM if the data size exceeds the threshold? A bigger question is is `makeShmBytes` is needed anywhere at all?

src/NodeSharedPrivate.hh
azeey, line 112: I learned that adding `zenoh::try_init_log_from_env();` here can be useful for debugging. You can then run `RUST_LOG=zenoh_shm=trace GZ_TRANSPORT_IMPLEMENTATION=zenoh gz sim camera_sensor.sdf` see trace message from the Zenoh shm submodule. This was an easy way for me to check that SHM is actually being used.

I haven't researched whether we can leave that in the code permannently. This is just a tip for debugging.

src/NodeShared.cc
```diff
@@ -1952,6 +1963,87 @@ std::shared_ptr<zenoh::Session> NodeShared::Session()
...
+    info.SetType(msgType);
+
+    HandlerInfo handlerInfo = this->CheckHandlerInfo(_topic);
+
+    // SHM-optimized receive: dispatch from a contiguous view into the SHM
+    // buffer when available, falling back to a copied string otherwise.
+    withPayloadView(_sample.get_payload(),
+      [&](const char *_data, std::size_t _size)
+      {
+        this->TriggerCallbacks(info, _data, _size, handlerInfo);
```
azeey, line 2004: In a follow up, we should take a look at triggering the callbacks from a separate thread. This relates to https://github.com/gazebosim/gz-transport/issues/955. I believe the Zenoh thread is blocked here while the callbacks are running.

include/gz/transport/detail/Node.hh
caguero: Thanks, reverted in ba07dbca.

src/Node.cc
caguero: Updated in f1c25c4e. The SHM branch itself is gone as of f711a625, so PublishRaw is back to a plain put().

include/gz/transport/SubscriptionHandler.hh
caguero: I'd rather merge this as is so the backport stays clean. I'll open a follow-up PR on main that makes CreateMsgFromBuffer virtual and drops the factory.

src/ReqHandler.cc
caguero: Good point and I measured it. With Zenoh's transport optimization doing the copy, service latency stays within a few percent up to 2 MB, so the explicit copy isn't worth it. `makeShmBytes` removed in f711a625. I'd preserve the direct-to-SHM serialization in Publish(). This is a measurable win (~20% at 1-2 MB).

src/NodeSharedPrivate.hh
caguero: Thanks for the tip. Added `zenoh::try_init_log_from_env() `before opening the session and documented RUST_LOG in the env var tutorial in 916dadc4. It's a no-op unless RUST_LOG is set, so it can stay.

src/NodeShared.cc
caguero: Agreed, we should tackle it as part of #955.

azeey (approved):

PR #848, opened 2026-04-10
Fix async signal safety undefined behavior in waitForShutdown()

caguero:
# 🦟 Bug fix

`waitForShutdown()` previously called `std::mutex::lock()`, `std::mutex::unlock()`, and `std::condition_variable::notify_all()` from inside its `SIGINT`/`SIGTERM` handler. Today I learned that none of these are on POSIX's async-signal-safe list (`signal-safety(7)`), so the behavior was undefined on Linux/macOS 🤯.
                                                                                                                                                        
This patch replaces the mutex/CV machinery with the **self-pipe trick**: the signal handler does only an async-signal-safe `write()` of one byte to a pipe, and `waitForShutdown()` blocks on `read()` of the read end. Instant wakeup, no polling, no locks, no undefined behavior.

## Checklist
- [x] Added tests

Generated-by: Claude Opus 4.6

### DIFF
--- a/src/WaitHelpers.cc
+++ b/src/WaitHelpers.cc
@@ -16,10 +16,17 @@
 */
 
 #include <algorithm>
-#include <condition_variable>
+#include <cerrno>
 #include <csignal>
-#include <mutex>
+#include <string>
 #include <vector>
+
+#ifdef _WIN32
+  #include <io.h>
+  #include <fcntl.h>
+#else
+  #include <unistd.h>
+#endif
 
 #include "gz/transport/WaitHelpers.hh"
 #include "gz/transport/Node.hh"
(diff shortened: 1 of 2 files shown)


iche033 (commented):
the signal handler in gz-common also has the self-pipe design (https://github.com/gazebosim/gz-common/pull/618). Not sure if it's worth switching to that or if you prefer to keep the version here.

caguero:
> the signal handler in gz-common also has the self-pipe design ([gazebosim/gz-common#618](https://github.com/gazebosim/gz-common/pull/618)). Not sure if it's worth switching to that or if you prefer to keep the version here.

Do you mean copying the code to keep the same implementation? I don't think we want to depend on gz-common.

iche033:
> > the signal handler in gz-common also has the self-pipe design ([gazebosim/gz-common#618](https://github.com/gazebosim/gz-common/pull/618)). Not sure if it's worth switching to that or if you prefer to keep the version here.
> 
> Do you mean copying the code to keep the same implementation? I don't think we want to depend on gz-common.

ah yes forgot gz-common is not a gz-transport dependency

iche033 (approved):

caguero:
@Mergifyio backport gz-transport15 gz-transport14 gz-transport13 ign-transport11

PR #867, opened 2026-05-14
Integrate Zenoh 1.8.0 - Part 1/2

### LINKED ISSUE #754: Zenoh outstanding tasks

caguero:
## Summary

This patch fixes different Zenoh teardown bugs that surfaced while testing Zenoh 1.8.0: use-after-free between tests, process crash at exit, and self-deadlock in service callbacks. Even Gazebo was crashing on shutdown.

Split into reviewable commits:

1. [7a13819](https://github.com/gazebosim/gz-transport/pull/867/commits/7a13819725f4c7ed1be05e15ab1c5b8dd3db6775): Add weak_ptr holder structs and ZenohTeardownEntity helper.
2. [043c1bc](https://github.com/gazebosim/gz-transport/pull/867/commits/043c1bc5a70cfe22a621f6eefac6463fcd97c161): Subscriber holder dispatch and detached undeclare on teardown.
3. [21b38c6](https://github.com/gazebosim/gz-transport/pull/867/commits/21b38c668a3a515e67b10b1a52cec6ab60bf033e): Queryable holder dispatch, detached undeclare, safe RunCallback default.
4. [a579173](https://github.com/gazebosim/gz-transport/pull/867/commits/a579173f11ae182883fb10cc990f29e08b5cf8c0): Publisher release teardown to avoid undeclare blocking at exit.
5. [0be5d86](https://github.com/gazebosim/gz-transport/pull/867/commits/0be5d86ee79d7ca61cc1436530c20ddd2306ddb0): Add `INTEGRATION_zenohShutdownStress` test.

## Checklist
- [x] Added tests

Generated-by: Opus 4.7


include/gz/transport/RepHandler.hh
```diff
@@ -82,8 +82,18 @@ namespace gz::transport
     /// callback function.
     /// \param[out] _rep Out parameter with the data serialized.
     /// \return Service call result.
+    /// \note Returns false by default. Derived classes override.
+    /// A safe default is provided (instead of leaving the method
+    /// abstract) because a Zenoh worker thread may invoke this while
```
azeey: Do we still need this chagnge since `RunCallback` is not called from the Zenoh handler?

include/gz/transport/RepHandler.hh
```diff
@@ -158,6 +178,19 @@ namespace gz::transport
...
+          Req req;
+          if (!req.ParseFromString(request))
+            return false;
+          Rep rep;
+          if (!_cb(req, rep))
+            return false;
+          if (!rep.SerializeToString(&response))
+            return false;
+          return true;
+        });
```
azeey, line 179: This duplicates code in `RunCallback`. I used Gemini to see if there's a different approach. The suggestion was to have `RepHandler` ( the derived class) to also inherit from `std::enable_shared_from_this<RepHandler<Req, Rep>>`. This would allow us to avoid creating the holder data structures and rely on weak pointers to a `RepHandler` instance. I have the suggested implementation in https://github.com/gazebosim/gz-transport/pull/new/azeey/refactor-zenoh-holders. Let me know what you think.

src/Node.cc
```diff
@@ -149,6 +154,42 @@ class Node::PublisherPrivate
...
+    if (!this->zenohIsShutdown.compare_exchange_strong(expected, true,
+          std::memory_order_acq_rel, std::memory_order_relaxed))
+      return;
+
+    // Deliberately leak the Zenoh wrappers. release() hands the raw
+    // pointer over to nobody and skips the wrapper's destructor,
+    // which would otherwise call undeclare() and crash or deadlock
+    // (see the doc comment above).
+    (void) this->zToken.release();
+    (void) this->zPub.release();
```
azeey, line 189: So, we're leaking these everytime a Publisher is destroyed? I can see in some setups (tests, benchmarks, RL trainings), we'd be creating and destroying publishers rapidly inside a single process. 

Maybe in a follow up PR, we could implement the following suggestion from Gemini:
> * **How it works:** Introduce a thread-safe cleanup queue inside `NodeShared` and a single dedicated background worker thread. When any Zenoh entity (Publisher, Subscriber, etc.) is destroyed, its `unique_ptr` is simply moved into this queue. The background thread pops entities off the queue and calls `undeclare()` safely.
> * **Pros:** Zero detached threads. No deadlocks. Highly performant for rapidly creating/destroying short-lived publishers.
> * **Cons:** Requires a slightly larger refactoring of `NodeShared` to manage the lifecycle of this cleanup thread and queue. When `NodeShared` itself shuts down, it can just drain the queue and intentionally discard the remaining Zenoh entities without calling `undeclare()`, cleanly avoiding the exit crash.


Can you create an issue to track this?

include/gz/transport/RepHandler.hh
caguero: No, we don't. It went back to pure virtual in 1656aa87.

include/gz/transport/RepHandler.hh
caguero: Thanks! I adopted your branch in 1656aa87 with two small tweaks: the generic subscriber keeps the explicit type filter and the missing attachment warning is back.

src/Node.cc
caguero: Thanks! I went for the fix 7b2fc1e7 directly in #868. It undeclares the publisher entities inline now. Publishers run no callbacks so there is nothing for undeclare to wait on, and the explicit `NodeShared::Shutdown()` added in that PR should make it safe at exit too. There is also a new INTEGRATION_zenohPublisherLeak test that fails against the old code.

include/gz/transport/RepHandler.hh
```diff
@@ -126,7 +135,8 @@ namespace gz::transport
   // the service call. 'Rep' is the protobuf message type that will be filled
   /// with the service response.
   template <typename Req, typename Rep> class RepHandler
-    : public IRepHandler
+    : public IRepHandler,
+      public std::enable_shared_from_this<RepHandler<Req, Rep>>
```
luca-della-vedova, line 139: I noticed that the other PR takes the `weak_ptr` from the interface class instead of the templated class, should we make both of them consistent?

caguero: It's intentional. In the srv request path the `weak_ptr` is created in `IReqHandler` code where the handlers are type erased and the reply only needs the virtual `NotifyResult()`, so the interface is the only possible place. In the subscriber/replier paths the `weak_ptr` is created inside the templated class and the dispatch needs typed calls.

luca-della-vedova:
Tried to look at how `rmw_zenoh` accomplishes this and opened https://github.com/gazebosim/gz-transport/pull/947 targeting this branch

luca-della-vedova (approved):

luca-della-vedova:
CI failure is a new compile warning, should be a straightforward fix

luca-della-vedova:
@caguero do you think this can be backported?

caguero:
> @caguero do you think this can be backported?

It should. I wonder if we should merge #868 first, let it soak a bit, and then backport both PRs to Jetty when we confirm nothing breaks.

caguero:
@mergifyio backport gz-transport15

PR #490, opened 2024-04-01
Revert "Use `std::shared_ptr` for `gz::transport::NodeShared` (#484)"

azeey:
# 🦟 Bug fix

## Summary

#484 fixed the `INTEGRATION_triggered_publisher` test in gz-sim8, but introduced a couple of regressions (see https://build.osrfoundation.org/job/gz_sim-ci-gz-sim8-jammy-amd64/82/ and https://build.osrfoundation.org/job/gz_sim-ci-gz-sim8-jammy-amd64/83/):

**NetworkHandshake.Updates**
```
Error Message
/home/jenkins/workspace/gz_sim-ci-gz-sim8-jammy-amd64/gz-sim/test/integration/network_handshake.cc:220
Expected: (100u) <= (zPos.size()), actual: 100 vs 70
```

and **ServerRepeat/SceneBroadcasterTest.StateStatic/0 (from ServerRepeat_SceneBroadcasterTest)**
```

Error Message
/home/jenkins/workspace/gz_sim-ci-gz-sim8-jammy-amd64/gz-sim/test/integration/scene_broadcaster_system.cc:588
Expected: (stepMsg.stats().real_time().nsec()) < (_msg.stats().real_time().nsec()), actual: 943661800 vs 44593094
```

It is possible to fix the regressions, but I did notice deadlocks while testing my fixes locally. The deadlocks occur as `NodeShared` was being destroyed, so I think it's best to revert this and investigate why the deadlock occurs. I'll open an issue to track it.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Rebase and Merge**

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

Issue #101, opened 2019-02-04
Remove NodeShared singleton


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-transport/issues/101)) by Nate Koenig (Bitbucket: [Nathan Koenig](https://bitbucket.org/%7Bc862cdd9-fcc8-4419-9fc7-e20db14b8fcb%7D/)).**

----------------------------------------
# Summary

NodeShared is a singleton, which can lead to runtime problems particularly during shutdown. One problem occurs when two nodes are in the same process, with one node publishing messages that the other node consumes and then publishes new messages. I see a segfault when running this code (https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-launch/commits/11b5830128da4561711a043c6450545a7b90c0a4 (ignitionrobotics/ign-launch@eea15cea322bc946f6f2a627e5ad8a54882bebff)) with this configuration file( https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-launch/commits/11b5830128da4561711a043c6450545a7b90c0a4 (ignitionrobotics/ign-launch@eea15cea322bc946f6f2a627e5ad8a54882bebff)#chg-config/gazebo.ign). When this version of `ign-launch` is installed, you can test using `ign launch -f config/gazebo.ign`. 

# Motivation

In addition to the above error, we are trying to remove singletons from our code base. For reasons see https://stackoverflow.com/questions/137975/what-is-so-bad-about-singletons/138012#138012 and https://cocoacasts.com/are-singletons-bad.

# Describe alternatives you've considered

I think @caguero mentioned that we could potentially use features built into ZeroMq to avoid the NodeShared singleton.

joxoby:
Related: https://github.com/ignitionrobotics/ign-transport/issues/137

PR #868, opened 2026-05-14
Integrate Zenoh 1.8.0 - Part 2/2

### LINKED ISSUE #754: Zenoh outstanding tasks

caguero:
## Summary

This patch completes the Zenoh 1.8 integration started in #867:

1. **Cold start race in service requests.** With interest driven routing (Zenoh > 1.6), the first `Node::Request` to a freshly spawned responder can time out silently. Requests now go through a per process cached `zenoh::Querier` whose interest declaration stays alive, so an in flight request reaches a queryable that appears late (`INTEGRATION_twoProcsSrvCallLateResponder`). A synchronous liveliness drain at `NodeShared` construction covers the topic side (it is also what fixes the `gz topic -i` and `gz service -i` cold start flakiness).

2. **Publisher and Querier cleanup.** Destroyed publishers are undeclared instead of leaked, and `INTEGRATION_zenohPublisherLeak` checks that they disappear from remote discovery right away. Cached Queriers live as long as the process, like `NodeShared` itself, which is never destroyed on purpose (see #101, #490 and #954). `Instance()` now documents that, and the teardown code in `~NodeShared` that never ran is gone.

3. **Async service requests.** `CreateZenohGet` no longer blocks while holding `NodeShared::mutex`: replies arrive through a weak_ptr and notify the waiting request, mirroring the ZeroMQ flow, and request handlers are removed from storage when the request ends.

The minimum Zenoh version is relaxed to 1.7.2 after verifying the API surface is identical (drop that commit if we prefer a 1.8.0 floor). The querier machinery follows the same simplification rules applied in #947: no intermediate holder structs, no setter plumbing.

Verified with the full test suite under both zeromq and zenoh (529 tests each) and by launching and interrupting `gz sim` repeatedly (clean exits, no panics).

## Checklist
- [x] Added tests

Generated-by: Claude Opus 4.7


src/NodeShared.cc
```diff
@@ -303,6 +303,69 @@ NodeShared::NodeShared()
...
+    try
+    {
+      zenoh::Session::LivelinessGetOptions opts =
+        zenoh::Session::LivelinessGetOptions::create_default();
+      opts.timeout_ms = kZenohLivelinessGetTimeoutMs;
+
+      zenoh::ZResult result = Z_OK;
+      auto replies = this->Session()->liveliness_get(
+        zenoh::KeyExpr("@gz/**"),
+        zenoh::channels::FifoChannel(SIZE_MAX - 1),
```
luca-della-vedova: I think this won't work, I tried to run this and traced a segfault to this line.
It seems what it tries to do is allocate a channel with a capacity of `size_t-1` which is incredibly large, the allocation fails and the code segfaults.
This is wildly out of my comfort zone so I had Gemini suggest a fix that seems to simplify the code a fair bit using an API with a lambda instead of a channel that could grow indefinitely.
This _seems_ to fix the segfault for me but again take it with a lot of salt since I am not familiar with Zenoh or gz-transport:

```patch
diff --git a/src/NodeShared.cc b/src/NodeShared.cc
index 2d174850..2b44b609 100644
--- a/src/NodeShared.cc
+++ b/src/NodeShared.cc
@@ -338,28 +338,20 @@ NodeShared::NodeShared()
       opts.timeout_ms = kZenohLivelinessGetTimeoutMs;
 
       zenoh::ZResult result = Z_OK;
-      auto replies = this->Session()->liveliness_get(
+      this->Session()->liveliness_get(
         zenoh::KeyExpr("@gz/**"),
-        zenoh::channels::FifoChannel(SIZE_MAX - 1),
+        [this](zenoh::Reply &reply)
+        {
+          if (reply.is_ok())
+          {
+            const auto &sample = reply.get_ok();
+            this->dataPtr->msgDiscovery->LivelinessMsgDataHandler(sample);
+            this->dataPtr->srvDiscovery->LivelinessSrvDataHandler(sample);
+          }
+        },
+        []() {},
         std::move(opts),
         &result);
-
-      if (result == Z_OK)
-      {
-        for (auto res = replies.recv();
-             std::holds_alternative<zenoh::Reply>(res);
-             res = replies.recv())
-        {
-          const auto &reply = std::get<zenoh::Reply>(res);
-          if (!reply.is_ok())
-            continue;
-          const auto &sample = reply.get_ok();
-          // Both handlers filter by entityType internally, so it is
-          // safe to dispatch every sample to both.
-          this->dataPtr->msgDiscovery->LivelinessMsgDataHandler(sample);
-          this->dataPtr->srvDiscovery->LivelinessSrvDataHandler(sample);
-        }
-      }
     }
     catch (const zenoh::ZException &e)
     {
```

caguero: > I think this won't work, I tried to run this and traced a segfault to this line. It seems what it tries to do is allocate a channel with a capacity of `size_t-1` which is incredibly large, the allocation fails and the code segfaults. This is wildly out of my comfort zone so I had Gemini suggest a fix that seems to simplify the code a fair bit using an API with a lambda instead of a channel that could grow indefinitely. This _seems_ to fix the segfault for me but again take it with a lot of salt since I am not familiar with Zenoh or gz-transport:


Thanks!

I could not reproduce the segfault here. Did it happen in gz-transport or testing with gz-sim or something else? I thought that the `FifoChannel` was doing only as a logical bound, not an actual preallocation. And just to confirm, are you testing with Zenoh 1.8.0 or a different version?
  
That said, I'm testing the callback approach with one latch. Otherwise, your suggestion is fire and forget and reintroduces the cold start race this PR is trying to fix.

luca-della-vedova: I'm on 1.7.2, I was testing with the CLI (i.e. `gz topic -l`).
Yea again can't really tell what is happening.  I dug a bit deeper and indeed you are right and it doesn't preallocate and it was some sort of hallucination, now it's pointing to some sort of double free? Again not sure but the callback based method seems to work.
How have you been testing this? I have been building zenoh-c / zenoh-cpp / zenoh from source on the 1.7.2 tag in a colcon workspace

caguero: I'm testing running all tests in gz-transport, and then, running gz-sim, and making sure that it doesn't crash when shutting down. I'm on Zenoh 1.8.0 right now but 1.7.2 should be fine as well.

luca-della-vedova: Actually hem apologies, I tried again to do a full clean build and actually it was working fine? I'll try to stress test this set of PRs a bit more but in general functionally it looks good

cheriehu:
I’ve tried some functional tests on this branch with a few of the example worlds and things seem to work as expected. This is what I’ve tried. 
1. Sending cmd_vel via the command line on diff_drive.sdf. The robot moves. 
2. Pausing, resuming, and resetting the simulation via a gz service call from the command line works
3. In sensors_demo.sdf, started the server and GUI in separate terminals. Echoed some camera topics, there was data being published. 
4. Started a world with server and GUI and separate terminals. Spawned a box from the GUI and saw a debug message on the server terminal reflecting the change. Calling the`/world/world_name/create` gz service to spawn a box also led to the box being spawned in the GUI
5. Tried `gz log record` and  `gz log playback` and verified that the playback looks right (at least for the /stats topic)

src/NodeShared.cc
```diff
@@ -336,6 +381,36 @@ NodeShared::~NodeShared()
...
+    {
+      this->dataPtr->session->close();
+    }
+    catch (const zenoh::ZException &e)
+    {
+      std::cerr << "gz-transport zenoh: session close failed: "
+                << e.what() << "\n";
+    }
+  }
+#endif
```
luca-della-vedova: I found out while running this that _I believe_ this code is never actually executed? Ticketed #954 for more details, but I might have tested this wrongly

caguero: Good catch, thanks for opening #954. You are right, `Instance()` stores a raw `new NodeShared` in a static map and nothing deletes it, so `~NodeShared` never runs. Removed in e73f2cbe, and `Instance()` now documents the leak.

One data point for #954: I prototyped deleting the instance from a static destructor and Zenoh aborts at exit on purpose ("Calling the Zenoh API at process exit is not supported"), because glibc runs thread local destructors before any static or atexit destructor. The only approach that works is an explicit shutdown call before main returns, the way rmw_zenoh does it, so I would rather tackle it in a separate PR if you agree.

include/gz/transport/detail/Node.hh
```diff
@@ -655,6 +655,17 @@ namespace gz::transport
...
+    // This request is finished (answered or timed out): remove the
+    // handler from the requests storage. On the Zenoh path nothing
+    // else removes it (replies notify the handler directly through a
+    // weak_ptr, which a late reply then finds expired). On the
+    // ZeroMQ path this is a no-op after a reply (RecvSrvResponse
+    // already removed it) and fixes the storage leak of timed-out
+    // handlers; a late ZeroMQ reply is simply dropped.
+    this->Shared()->Requests().RemoveHandler(
+      fullyQualifiedTopic, this->NodeUuid(),
+      reqHandlerPtr->HandlerUuid());
```
luca-della-vedova, line 666: There is another `Request` path above that passes a function (so the asynchronous path), should there be a `RemoveHandler` there as well?

test/integration/CMakeLists.txt
```diff
@@ -19,6 +20,7 @@ set(tests
 
 if (HAVE_ZENOH)
   list(APPEND tests
+    zenohPublisherLeak.cc
```
luca-della-vedova, line 23: We should add these to the Bazel tests as well so we have parity.

luca-della-vedova (commented):
Yea I guess it's OK to do in a followup, the `Recorder` test failed in the last commit segfaulting at teardown again which makes me think we need to fix zenoh teardown.
A few minor comments, but again not too familiar with gz-transport so can't give an in-depth review

include/gz/transport/detail/Node.hh
caguero: Yes, it should. On Zenoh nothing removed the handler of an asynchronous request, so it stayed in the storage forever. Fixed in d45352ed: the handler is removed when the query finishes, and INTEGRATION_twoProcsSrvCall now checks that.

test/integration/CMakeLists.txt
caguero: Done in caf007b5.

caguero:
Thanks for the suggestions, both inline comments should be addressed.

The Recorder crash should be fixed in d0e74695. The Bazel job runs Zenoh 1.7.2, where undeclaring a subscriber does not wait for a callback that is still running. The Recorder has its own discovery object, so a liveliness callback could run after the Recorder was destroyed. Discovery now blocks that. Could you confirm on your side?

luca-della-vedova (approved):
Tentative approve based on my testing, but again not really my area of expertise

PR #915, opened 2026-07-30
Fix TopicList() subscribers information

### LINKED ISSUE #914: TopicList() may return incomplete topic information
## Environment
* OS Version: Ubuntu 24.04.4 LTS
* Source or binary build?
  * Source build
  * Branch: main
  * gz-transport version: latest commit
* ROS 2 Version: Rolling
* Gazebo Sim Version: v11.0.0~pre1

## Description
* Expected behavior: 
`gz::transport::Node::TopicList()` should return all available Gazebo Transport topics after the simulation has started, including topics that only have subscribers.
 
* Actual behavior: 
`gz::transport::Node::TopicList()`  always returns an incomplete topic list.

### Setup
A minimal reproduction package is provided here: https://github.com/azeey/gsoc2026_multirobot/tree/main/bridge/topiclist_test

The package contains a simple ROS 2 node (`topiclist_test_node`) that:

* Creates a Gazebo Transport node.
* Periodically calls `gz::transport::Node::TopicList()`.
* Compares the returned topic list with a predefined list of expected Gazebo topics.
* Reports missing topics in the console output.

The node supports the following launch parameters:

* `use_composition`: whether to load the node as a composable node.
  * Default: `True`
* `freq`: the frequency (Hz) at which `TopicList()` is called.
  * Default: `1.0`

### Steps
1. Build & Source
``` bash
colcon build --symlink-install
source install/setup.bash
```
2. Launch
``` bash
ros2 launch topiclist_test topiclist_test.launch.py use_composition:=True freq:=1.0
```
3. Observe the output
The node periodically prints the number of expected topics, the number of topics returned by `TopicList()`, and any missing topics.

### Output
The test was performed with different `TopicList()` query frequencies (0.1 Hz, 1 Hz, and 2 Hz) and with both composable and non-composable node modes.

In all cases, the same incomplete topic list was consistently returned. The node was running for more than 2 minutes, but the missing topics did not appear in the result.
``` bash
[INFO] [topiclist_test_node]: Expected Gazebo topics: 25, discovered by TopicList(): 21

[INFO] [topiclist_test_node]: Missing topics:
  /model/vehicle/cmd_vel
  /model/vehicle/enable
  /world/test_world/light_config
  /world/test_world/material_color
```

### 1. Possible cause: asynchronous discovery

The issue appears to be related to the asynchronous nature of Gazebo Transport discovery.

From the current implementation, `TopicList()` requests subscriber information and then immediately reads the cached discovery information. Since discovery responses are processed asynchronously, the cache may not contain all remote subscriber information when the topic list is generated.

### 2. Comparison with `gz topic --list`

Interestingly, the command line tool `gz topic --list` does not show this issue. Under the same simulation setup, it returns the complete topic list, including the topics missing from `TopicList()`.

This suggests that the underlying discovery mechanism can eventually obtain the complete topic information, but `TopicList()` may return before all discovery responses have been received.

### 3. Experiment: waiting for discovery responses

Adding a small delay inside `TopicList()` before reading the discovery cache improves the completeness of the result. For example, adding a sleep after requesting subscriber information in [Discovery::TopicList()] (https://github.com/gazebosim/gz-transport/blob/0fdc2dfab785dde5ca0bfd9c989de5783b44afc5/src/Discovery.hh#L818-L846):

```cpp
      public: void TopicList(std::vector<std::string> &_topics)
      {
        ...
        // Request the list of subscribers.
        Publisher pub("", "", this->pUuid, "", AdvertiseOptions());
        this->SendMsg(
          DestinationType::ALL, msgs::Discovery::SUBSCRIBERS_REQ, pub);

        this->WaitForInit();

        std::this_thread::sleep_for(std::chrono::milliseconds(1));

        std::lock_guard<std::mutex> lock(this->mutex);
        this->info.TopicList(_topics);
        ...
    }
```
With this modification, `TopicList()` returns the complete topic list in the current reproduction case.

However, the required waiting time depends on the number of topics and the discovery state. In a scenario with a larger number of topics, a longer delay is required before `TopicList()` can return the complete topic list.

### 4. Local process discovery limitation
Additionally, another issue was observed when using `use_composition:=True`. The discovery logic ignores information from local processes, so topics subscribed by nodes running in the same process are not included in the result. Therefore, when the test node is loaded as a composable node in the same process, TopicList() cannot discover topics from that local process.
https://github.com/gazebosim/gz-transport/blob/0fdc2dfab785dde5ca0bfd9c989de5783b44afc5/src/Discovery.hh#L1123-L1142

caguero:
# 🦟 Bug fix

Partially fixes #914

## Summary

The patch fixes the issue by not clearing the list of remote subscribers every time we call `TopicList()`. An additional fix is that we also add the local subscribers, as local discovery messages are ignored. And finally, we won't send anymore the `SUBSCRIBERS_REQ` message if we're in the service discovery instance.

**Note**: Right now, the `SUBSCRIBERS_REQ` is non-blocking (as before), so we don't wait a specific time to get the subscribers. Now, each subscriber received is preserved for the future (unless it unsubscribes), so future TopicList has higher change to get the complete list of subscribers. Before this patch it was hard, as we were clearing the subscribers each time. But we still don't have a consistent way to get the the subscribers during the first call.

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests

Assisted-by: Claude Fable 5


C88-YQ (commented):
LGTM!!

C88-YQ:
@caguero Thanks for the PR. I also do some behavior tests to verify the expected behavior.

## Test setup
I used my own test package:

https://github.com/azeey/gsoc2026_multirobot/tree/main/bridge/topiclist_test

The package provides a simple ROS 2 node (`topiclist_test_node`) that:

* Creates a Gazebo Transport node.
* Periodically calls `gz::transport::Node::TopicList()` and `gz::transport::Node::ServiceList()`.
* Compares the returned lists with predefined expected topics/services.
* Reports missing or extra topics/services.

It also includes a [bash script](https://github.com/azeey/gsoc2026_multirobot/blob/main/bridge/topiclist_test/check_gz_lists.sh) that periodically runs `gz topic --list` and `gz service --list` and collects the results.

## Test Result

I tested the behavior with both Gazebo Transport middleware backends: UDP and Zenoh. The results are summarized below.

### UDP

1. The results are the same whether `use_composition` is set to `true` or `false`.
2. Both `gz topic --list` and `gz service --list` consistently return the complete list of topics/services. There is still a small delay before updates appear, which seems expected since the behavior was the same before this change.
3. Both command line tools and API calls from another node return correct results. I did not observe any missing or extra topics/services.
4. The updates are very responsive. After adding or removing an entity, the list is updated correctly within the same cycle.

### Zenoh

1. The results are the same whether `use_composition` is set to `true` or `false`.
2. Calling `TopicList()` and `ServiceList()` from another node returns correct results. Updates are also very responsive.
3. `gz topic --list` returns quickly and always gives the correct result.
4. However, `gz service --list` does not work reliably. In 60 tests, only 3 returned the complete list, while the others missed different services. I also tested this on the `main` branch and observed the same behavior, so this does not appear to be introduced by this PR. I will investigate this issue separately.

Overall, LGTM. Thanks for the fix!

C88-YQ (approved):

caguero:
@mergifyio backport gz-transport15 gz-transport14 gz-transport13

PR #928, opened 2026-08-18
Fix unbounded memory growth with slow local subscribers

### LINKED ISSUE #926: Unbounded memory growth with local subscribers and slow callbacks
When doing pub sub with local subscriptions, whenever a message is published, we push it to a `std::list` https://github.com/gazebosim/gz-transport/blob/50066909f293f5c016c9a28986f36553700a86a8/src/Node.cc#L524-L525
The list is handled in a [separate thread](https://github.com/gazebosim/gz-transport/blob/50066909f293f5c016c9a28986f36553700a86a8/src/NodeShared.cc#L1653) that continuously:

* [Pops a message](https://github.com/gazebosim/gz-transport/blob/50066909f293f5c016c9a28986f36553700a86a8/src/NodeShared.cc#L1681) from the front of the queue.
* [Executes all the callbacks](https://github.com/gazebosim/gz-transport/blob/50066909f293f5c016c9a28986f36553700a86a8/src/NodeShared.cc#L1685-L1699) one by one.

This is potentially problematic because if the callbacks don't process data fast enough the queue will grow in an unbounded way, potentially killing off the applications.

## Reproduce it!

I observed this happening when having images being published and a local callback that had to do somewhat heavy lifting and AI generated a reproduction example.

https://github.com/luca-della-vedova/gz_transport_leak_repro

The example publishes high frequency camera data and has an artificially slow callback that prints the memory consumption, you can see how it will grow indefinitely.

## Possible solutions

I very naively added a check [where we push](https://github.com/gazebosim/gz-transport/blob/50066909f293f5c016c9a28986f36553700a86a8/src/Node.cc#L524-L525) into the publisher queue of "if queue size > N, pop_front from the queue" and that fixed the leak.
I suppose other middlewares (i.e. ROS) have a queue size parameter for both publishers and subscribers, but it seems gz-transport doesn't, should we consider introducing it or just have a "large safe value" to avoid this kind of issues killing applications?
https://github.com/gazebosim/gz-transport/issues/24 is probably related (a queue size would pretty much implement a circular buffer).

caguero:
# 🦟 Bug fix

Fixes #926

## Summary

The queue delivering messages to local (intraprocess) subscribers was unbounded, so callbacks slower than the publication rate grew it until out of memory. This patch bounds it per topic: when a topic already has `GZ_TRANSPORT_LOCAL_HWM` messages queued (default 1000), publishing a new one drops that topic's oldest queued message and warns once per overload episode. Setting the variable to `0` restores the old unbounded behavior. Mirrors
  `GZ_TRANSPORT_RCVHWM`/`SNDHWM` and works on both backends.

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [x] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [x] Added tests
- [x] Updated migration guide (as needed)

Assisted-by: Fable 5

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -14,6 +14,19 @@ release will remove the deprecated code.
     * `waitForShutdown()` moved from `Node.hh` / `Node.cc` to `WaitHelpers.hh` / `WaitHelpers.cc`. `Node.hh` re-exports via `#include "gz/transport/WaitHelpers.hh"`, so no code changes are required for existing users of `waitForShutdown()`.
     * `waitUntil()`, `waitForService()`, `waitForTopic()` moved from `Helpers.hh` / `Helpers.cc` to `WaitHelpers.hh` / `WaitHelpers.cc`. Code using these functions must now `#include "gz/transport/WaitHelpers.hh"` directly.
 
+### Modifications
+
+1. The queue used to deliver messages to local (intraprocess) subscribers
+   is now bounded per topic. When a topic has `GZ_TRANSPORT_LOCAL_HWM`
+   messages queued (1000 by default), publishing a new message drops the
+   oldest queued message of that topic instead of letting the queue grow
+   without limit. Previously the queue was unbounded, so a subscriber
+   processing messages slower than the publication rate would make the
+   process run out of memory. Set the `GZ_TRANSPORT_LOCAL_HWM` environment
+   variable to tune the capacity, or to 0 to restore the previous unbounded
+   behavior. The new `localHwm()` function returns the current capacity.
+   * [GitHub issue 926](https://github.com/gazebosim/gz-transport/issues/926)
+
 ### Deprecations
 
 1. The `gzTransportPublish` function in `CIface.h` has been deprecated because
--- a/include/gz/transport/Node.hh
+++ b/include/gz/transport/Node.hh
@@ -73,6 +73,19 @@ namespace gz::transport
   /// If your buffer reaches the maximum capacity data will be dropped.
   int GZ_TRANSPORT_VISIBLE sndHwm();
 
+  /// \brief Get the capacity (High Water Mark) of the queue that stores
+  /// messages to be delivered to local (intraprocess) subscribers. Note
+  /// that this limit is applied per topic.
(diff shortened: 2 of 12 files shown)


luca-della-vedova (commented):
I'm not familiar with the history of `gz-transport` but is there a reason we should still commit to this strategy of using a global parameter set through environment variables for queue sizes, now with a new parameter for local queues that is different from the publish / subscribe queue size?

IMHO ROS does it in a sensible way that users are used to, with a publisher queue / subscriber queue parameter for each publisher and subscriber, and no distinction between local and remote pub/sub queue sizes.
Apart from converging towards the way ROS does things, a global parameter is not a great approach imho. Different topics might have very different requirements for queue sizes. For example users might want a larger queue size for data they don't want to drop, or a smaller queue size (as small as 1) for data that they will just parse once in a while and they only want the latest sample.

caguero:
> I'm not familiar with the history of `gz-transport` but is there a reason we should still commit to this strategy of using a global parameter set through environment variables for queue sizes, now with a new parameter for local queues that is different from the publish / subscribe queue size?
> 
> IMHO ROS does it in a sensible way that users are used to, with a publisher queue / subscriber queue parameter for each publisher and subscriber, and no distinction between local and remote pub/sub queue sizes. Apart from converging towards the way ROS does things, a global parameter is not a great approach imho. Different topics might have very different requirements for queue sizes. For example users might want a larger queue size for data they don't want to drop, or a smaller queue size (as small as 1) for data that they will just parse once in a while and they only want the latest sample.

Thanks for the suggestion! I agree that this is the way to go in the medium/long-term but in the short term we're trying to address the potential out of memory crash. The global local env variable decision is there because it's the logical continuation of how it started in ZeroMQ exposing global env vars as well. Creating queues for publishers/subscribers at the topic level with a programmatic API will need more architecture and not sure if it'll be all backportable. We'll need to think about it and prototype. I'll keep working on it in the background in a separate PR.

PR #937, opened 2026-08-21
Add support for Zenoh in Bazel builds

luca-della-vedova:
# 🎉 New feature

## Summary

As per title, there was no Zenoh support for Bazel builds, also adds a CI job to verify that unit tests pass.
The most important (and potentially controversial) change is that the added dependency to zenoh is to *1.7.2* while the codebase was (I believe) targeting 1.5.0.
I tried this out and it seems to work (also can reference green CI in this PR), I referenced the `rmw_zenoh` bump PRs where indeed going from 1.5.0 to 1.7.1 (there was no 1.7.2 bump) was very straightforward and involved no code change:

* [1.6.1](https://github.com/ros2/rmw_zenoh/pull/834/changes), no diff, only bump dependency tags.
* [1.6.2](https://github.com/ros2/rmw_zenoh/pull/842/changes), Same.
* [1.6.2](https://github.com/ros2/rmw_zenoh/pull/870/changes), Same, with a new config that is not used.

By contrast the [1.8.0 bump](https://github.com/ros2/rmw_zenoh/pull/935) was a lot trickier, so tricky that it had to be reverted and done twice!

A rough overview of other changes:

* Add a `enable_zenoh` bazel flag to the build.
* The `enable_zenoh` actually changes the value of the `GZ_TRANSPORT_DEFAULT_IMPLEMENTATION` variable. This might not be strictly necessary but it's a bit confusing when users build gz-transport with `enable_zenoh:=true` and it doesn't actually use Zenoh unless they also set an environment variable.
* The various changes in the test BUILD.bazel are due to the fact that there are now dependencies on test config files, so they are needed for the tests

## Backport Policy
- [ ] This is safe to backport to the following versions:
    - [x] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure
## Test it

In CI, or locally with:

```
# With Zenoh
$ bazel test --//:enable_zenoh=true --test_output=errors //...
# Without zenoh, can also omit, or set enable_zenoh to false
$ bazel test --test_output=errors //...
```

## Checklist
- [x] Added tests

Assisted-by: Gemini


luca-della-vedova:
Discussed at Gazebo Asia community meeting that this seems to be OK and shouldn't break anything

BUILD.bazel
```diff
@@ -66,9 +82,14 @@ cc_library(
     copts = [
         "-Wno-deprecated-declarations",
     ],
-    defines = [
-        'GZ_TRANSPORT_DEFAULT_IMPLEMENTATION=\\"zeromq\\"',
-    ],
+    defines = select({
+        ":zenoh_enabled": [
+            'GZ_TRANSPORT_DEFAULT_IMPLEMENTATION=\\"zenoh\\"',
```
caguero: This makes a Bazel build with `enable_zenoh=true` default to zenoh at runtime, while a CMake build with zenoh enabled still defaults to zeromq (`GZ_TRANSPORT_DEFAULT_IMPLEMENTATION` is an independent cache variable there). I'd keep zeromq as the default in both build systems so they behave the same, and run the zenoh CI job with `--test_env=GZ_TRANSPORT_IMPLEMENTATION=zenoh` instead. Same test coverage, less code.

MODULE.bazel
```diff
@@ -13,6 +13,8 @@ bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "rules_proto", version = "7.1.0")
 bazel_dep(name = "sqlite3", version = "3.53.2")
 bazel_dep(name = "rules_cc", version = "0.1.2")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "zenoh-cpp", version = "1.7.2.bcr.1")
```
caguero, line 19: How about a small comment to keep all Zenoh min versions in sync:

```suggestion
# Minimum version (bzlmod may resolve higher). Keep in sync with
# GZ_TRANSPORT_ZENOH_MIN_VERSION in CMakeLists.txt (see #868).
bazel_dep(name = "zenoh-cpp", version = "1.7.2.bcr.1")
```

.github/workflows/bazel.yml
```diff
@@ -25,4 +26,21 @@ jobs:
       exclude_windows: true
       # Use opt compilation mode to prevent perf-related test flakiness.
       # This only adds a couple minutes of build time overhead.
-      bazel_test_command: "bazel test -c opt --test_output=errors //..."
+      bazel_test_command: "bazel test -c opt --//:enable_zenoh=false --test_output=errors //..."
+
+  test-zenoh:
```
caguero: Optional: both jobs only differ in the flag value. A `strategy.matrix` over `enable_zenoh` would avoid the duplication.

caguero (commented):
Looking good overall! I just left a few minor comments.

MODULE.bazel
luca-della-vedova: https://github.com/gazebosim/gz-transport/pull/937/commits/9058ccfb6505782eeffe7b2374e97d6c2c89906b

.github/workflows/bazel.yml
luca-della-vedova: Ah yes, nice! https://github.com/gazebosim/gz-transport/pull/937/commits/cb4209fe221908c8aec50665691316c70c2cc4a1

BUILD.bazel
luca-della-vedova: I just realized actually, is the purpose of the `HAVE_ZENOH` just to make gz-transport work regardless of whether zenoh is available or not and, in the future, if Zenoh is widely distributed we can just remove it and compile the support altogether?
If so, I just reverted this change, removed the `zenoh_enabled` condition altogether in https://github.com/gazebosim/gz-transport/pull/937/commits/5a17f7bdec89a07cc9715f7fec2d1329ce9ea4e4 and did your suggestion of using `test_env` to set the environment variable, since Zenoh is already available in BCR it will always be built (hardcode `HAVE_ZENOH` to true in Bazel builds) and users just need to specify the environment variable to choose their transport implementation.
This will cause generally longer build times for `gz-transport` but at least users don't need to mess with build flags (it can be a bit tricky to keep track of all the build flags needed for every downstream dependency when building large projects).

caguero: `HAVE_ZENOH` is a "was Zenoh found at configure time" flag that conditionally compiles the Zenoh code. Eventually, Gazebo M will set Zenoh the default backend. Then, we'll make it a required dependency and drop the conditionals, but I'd keep them in the code until then.
 
For Bazel, hardcoding `HAVE_ZENOH=1` is fine with me. No need for an `enable_zenoh` flag as users can pick the backend at runtime.

luca-della-vedova: Yap this is what I ended up doing in this PR, it hardcodes `HAVE_ZENOH` to true

caguero (approved):

luca-della-vedova:
@mergifyio backport gz-transport15

PR #947, opened 2026-08-26
Simplify zenoh shutdown logic

luca-della-vedova:
# 🎉 New feature

Simplify #867 by using `shared_from_this` and removing intermediate `std::function` objects, parallel cleanup threads in favor of pure RAII.

## Summary

To be honest I'm not very familiar with how Zenoh works but I thought we could take inspiration from how `rmw_zenoh` does the same thing since it's a library that is fairly tested and used by now.
Specifically, it seems the pattern used there is to use `enable_shared_from_this` for subscriptions, then pass a weak pointer to the lambda directly.
While the object is alive, the weak pointer `lock()` call will work as intended. As soon as the object is destroyed `lock()` will return a `nullptr` and we won't execute the lambdas.

Ref in rmw_zenoh https://github.com/ros2/rmw_zenoh/blob/e3159856e3816e726f2c5f1f9b4858ad6b7ee63e/rmw_zenoh_cpp/src/detail/rmw_subscription_data.cpp#L758-L764.

This seems to make the code quite a bit simpler, removing 200 LOC or so, all the new `std::function` wrappers, the parallel thread cleanup.

I also added missing files to the test bazel build (otherwise this PR would have made bazel red post https://github.com/gazebosim/gz-transport/pull/937)

## Checklist
- [ ] Added tests

Assisted-by: Gemini

## Test

Tested it in CI in my fork https://github.com/luca-della-vedova/gz-transport/pull/1

### DIFF
--- a/include/gz/transport/RepHandler.hh
+++ b/include/gz/transport/RepHandler.hh
@@ -59,6 +59,7 @@ namespace gz::transport
   /// \class IRepHandler RepHandler.hh gz/transport/RepHandler.hh
   /// \brief Interface class used to manage a replier handler.
   class GZ_TRANSPORT_VISIBLE IRepHandler
+    : public std::enable_shared_from_this<IRepHandler>
   {
     /// \brief Constructor.
     /// \param[in] _pUuid Process UUID.
@@ -98,15 +99,6 @@ namespace gz::transport
     public: virtual std::string RepTypeName() const = 0;
 
 #ifdef HAVE_ZENOH
-    /// \brief Temporarily store the dispatch closure. Must be called by the
-    /// most derived RepHandler subclass before CreateZenohQueriable.
-    /// The dispatch closure should capture a weak_ptr to the handler to safely
-    /// abort if invoked during handler destruction.
-    protected: void SetZenohQueryableDispatch(
-      const std::string &_service,
-      std::function<bool(const std::string &request,
-                         std::string &response)> _dispatch);
-
     /// \brief Create a Zenoh queriable.
     /// \param[in] _session Zenoh session.
     /// \param[in] _service The service.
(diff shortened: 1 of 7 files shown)


caguero (commented):
Thanks for the idea!

I like the weak pointer at the interface level, removing the `std::function` plumbing, the publisher RAII and the bazel fixes.

Removing the detached undeclare threads is still a problem. In zenoh 1.8, undeclare of a subscriber or queryable waits for its callback closure. With the weak pointer, the last handler reference often dies inside the callback of the same entity (for example a service callback that unadvertises its own service). The zenoh thread then waits for itself forever. Everything looks fine, but new service requests in that process time out.

To confirm this I added `INTEGRATION_zenohTeardownFromCallback` to #867 and merged it into yours, so your CI will show the failure now.

luca-della-vedova:
Ah I see, I have actually been testing with 1.5.0 (locally and CI since I don't have nightlies enabled) and 1.7.2 (Bazel), not 1.8.0 yet, must be why haven't bumped into this. I ran the new `INTEGRATION_zenohTeardownFromCallback` from a machine with the 1.5.0 apt package distributed and it still seems to work.

I tried a simple-ish thread fix in https://github.com/gazebosim/gz-transport/pull/947/commits/bac29825f5a6c285d836d9d8939f39247a07bfe3.
From what I understand, this avoids the deadlock since it just runs the pointer reset logic in a detached thread, and it also still uses pure RAII by just resetting the pointer (so it will be deallocated autonomously by its destructor) instead of leaking it manually.
It seems to make the test green again but not sure I missed anything.

caguero (approved):
Good to go now. Good iteration, thanks for the patch!

PR #952, opened 2026-09-01
Simplify zenoh 1.8.0 part 2

luca-della-vedova:
I used an agent to check how `rmw_zenoh` did its Zenoh implementation / 1.8.0 bump, and it seems that indeed they did it with a [channel](https://github.com/ros2/rmw_zenoh/blob/e754341cadd92c39f4f9d44908fa6714ff0eaa4c/rmw_zenoh_cpp/src/detail/rmw_context_impl_s.cpp#L474-L478) rather than a lambda callback, so this revert the change that I requested in the first place (apologies!).
This means we can also remove the `isShutdown` variable, together with the `drainMutex` since they were only used for that lambda based approach.

Also removes the wait:

```
    // Give scouting a short window to establish the first transport
    // so the liveliness query below can reach remote peers. Times
    // out harmlessly when this process is alone.
    waitUntil([this] { return !this->Session()->get_peers_z_id().empty(); },
              kZenohPeerWait, std::chrono::milliseconds(5));
```

From what I understand the `liveliness_get` call below already has a timeout so it shouldn't be necessary to have additional ones, but a bit out of my depth in all these changes.

### DIFF
--- a/src/NodeShared.cc
+++ b/src/NodeShared.cc
@@ -298,35 +298,12 @@ NodeShared::NodeShared()
 #ifdef HAVE_ZENOH
   else if (impl == "zenoh")
   {
-    this->dataPtr->msgDiscovery->Start(this->Session(),
-      std::bind(&MsgDiscovery::LivelinessMsgDataHandler,
-            this->dataPtr->msgDiscovery.get(), std::placeholders::_1));
-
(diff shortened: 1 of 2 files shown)


caguero (approved):
Works for me, thanks!

PR #968, opened 2026-09-22
Drop enable_shared_from_this from the Zenoh handlers

caguero:
# 🦟 Bug fix

Needed by #966 (backport of #867 to gz-transport15).

## Summary

The Zenoh handlers added in #867 and #868 derive from `std::enable_shared_from_this` so their callbacks can hold a weak reference. That base adds a `weak_ptr` member to `ISubscriptionHandler`, `RawSubscriptionHandler`, `IRepHandler` and `IReqHandler`, which breaks ABI when backported to Jetty. The ABI checker on #966 is red for this reason.

This PR removes the bases. The subscriber and queryable callbacks now resolve their handler through the `NodeShared` storage. The request path receives the owning `shared_ptr` from `NodeShared`, which already holds it.

## Backport Policy
    - [x] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

Assisted-by: Claude Fable 5.1


azeey:
I think the `main` first development approach is to encourage us to land breaking changes on main with the cleanest and most optimal approach/design. If we need to create a work around for maintaining ABI, shouldn't that land on the branches that need it, and not in main?

caguero:
> I think the `main` first development approach is to encourage us to land breaking changes on main with the cleanest and most optimal approach/design. If we need to create a work around for maintaining ABI, shouldn't that land on the branches that need it, and not in main?

I totally agree and I can be convinced otherwise but in this case the benefit of not diverging between `main` and `gz-transport15` is worth the patch in my opinion.

